PR #48, opened 2024-09-27
changed library in readme example (ASAR --> asar)

sbreitbart-NOAA:
Changed library from ASAR to asar so example code worked if copy/pasted from readme into user console


Schiano-NOAA (approved):
Thanks for making this change. It has been stuck on my todo list

PR #53, opened 2024-10-01
Added code to add_theme.R to use the nmfs color palette in plots

sbreitbart-NOAA:


Schiano-NOAA (changes requested):
It looks like some dependency is missing somewhere for nmfspalette. I think it might need to be added into the description using `usethis::use_package('nmfspalette')` as well as added as a remote if it already isn.'t.

@Bai-Li-NOAA could you check this out? I think the tests are also failing due to some issue in the tests.

Bai-Li-NOAA:
@Schiano-NOAA and @sbreitbart-NOAA The GHA workflows failed on push but not on the pull request. I suspect this might be due to the feature branch being behind the main branch. I ran a rebase test [here](https://github.com/nmfs-ost/asar/tree/fix-gha), and all tests passed with no changes required to the actual code base. Would you like me to rebase the `access_plot_theme branch` and force push it?

I noticed that `tests/testthat/test-add_theme.R` is currently commented out. It would be helpful to update the tests and include some for the newly added code (e.g., usage of `{nmfspalette}`) before merging these changes into the main branch.

Schiano-NOAA (approved):

PR #54, opened 2024-10-01
Create working example and resources for users

Schiano-NOAA:
Changing to a working example in read me that mimic that of a current stock assessment. The files resulting from the example and rendered report need to be added as their own example in a folder

### Additional Files to Add:

- [x] produced folder structure with quarto files resulting from the create_template function
- [x] converted output (.csv) file from the convert_output function
- [x] associated .bib files from the example
- [x] rendered pdf of the example
- [x] reference document of the final public assessment report for petrale sole


Schiano-NOAA:
Suggest moving the example files to asar-demo and create a link to this repo for users to go to

Schiano-NOAA:
I added most of these components into the tutorial github page instead so users can refer to it there. There are multiple links in different places to navigate to these resources

PR #59, opened 2024-10-08
Reorder skeleton

sbreitbart-NOAA:
# What is the feature?

- Updating the report skeleton and templates to reflect the changes voted on by steering committee (as of Sept 2024).

- Updating the section templates' names and content
- Changing the report section names to include numbers, so they have a logical order.
- Updating the test associated with the create_template() function to reflect these changes.

I don't believe so.


Schiano-NOAA (approved):
@sbreitbart-NOAA These changes look good. Thanks for catching some of the instances that references the tables and figures doc. I am going to hash out why the checks are failing then hopefully we can merge into the dev branch

Schiano-NOAA:
@Bai-Li-NOAA Could you check these tests and see why they are failing? I have been trying to fix it, but it doesn't seem to work. It looks to be failing for `test-add_section.R`. When I do it locally manually it passes, but not when I run `devtools::test()`

It also looks to also be presenting warnings for test-create_template due to the examples not containing a directory for resdir, but that is OK because the warning should be present when resdir = NULL. I changed the tests to have "data" as the resdir, but the warning keeps presenting.

It is almost like my changes aren't being recognized by the tests. No rush on this...

Bai-Li-NOAA:
@Schiano-NOAA , sure thing. I can work on it Tuesday morning. 

I had a quick look at the [GHA log](https://github.com/nmfs-ost/asar/actions/runs/11292619998/job/31408980953#step:7:788), and it seems like the expected output has 10 items, but the actual [output from add_sections()](https://github.com/nmfs-ost/asar/blob/reorder-skeleton/tests/testthat/test-add_section.R#L3C3-L7C23) is only 6. Could you double-check the [exp_list](https://github.com/nmfs-ost/asar/blob/reorder-skeleton/tests/testthat/test-add_section.R#L9C3-L20C4) and make any adjustments if needed? Same for the other exp_list [here](https://github.com/nmfs-ost/asar/blob/reorder-skeleton/tests/testthat/test-add_section.R#L32C3-L46C4). Thank you!

Schiano-NOAA:
@Bai-Li-NOAA thanks bai! Yeah I have already fixed this (I believe). Due to the new section naming, assessment is repeated and so I had to unlist the elements in add_base_section, then re list them using as.list. Doing so, the tests passed when I ran the script but not in the console or online. I will double check that this is still happening

Bai-Li-NOAA:
@Schiano-NOAA , I was able to replicate the `devtools::test()` error locally. In addition, when running the code line by line, I got the following warning message:
``` r
sec_list <- add_base_section(c("introduction", "assessment", "discussion"))
#> Warning message:
In section_list[i] <- sec_file :
  number of items to replace is not a multiple of replacement length
```
Looks like the issue is that there are two `add_base_section()` functions: one in `R/utils.R` and one in `R/add_section.R`. The tests automatically pick the one from `R/utils.R`. Removing the one from `R/utils.R` should fix the issue.

Schiano-NOAA:
@Bai-Li-NOAA Thanks for much for checking that Bai and making that catch! I am going to remove the one from utils and hopefully it will fix this issue.

PR #60, opened 2024-10-11
Update fxn examples

sbreitbart-NOAA:
# What is the feature?
* Adding or updating examples for functions.
* Clarifying parameter documentation, where necessary.

* Adding or updating examples for functions, which are wrapped in `\dontrun{}`. 

* I don't think so.


R/add_child.R
```diff
@@ -1,13 +1,22 @@
...
-#'         Utilize the cat() function to implement into readable text.
+#' @return Formatting R chunk for child document to add section
+#' into the template/skeleton. Utilize the cat() function to
+#' implement into readable text.
 #' @export
 #'
-#' @examples add_child("test_quarto.qmd", label = "test_doc")
+#' @examples
+#'\dontrun{
+#' add_child(x = "test_quarto.qmd",
```
Schiano-NOAA, line 17: Is the function set up that you need to add the .qmd extension? I thought I made it so they didn't need that. If it's the latter, then the extension will get replicated

R/add_section.R
```diff
@@ -11,11 +11,15 @@
...
 #' @export
 #'
-#' @examples add_section(
-#'   new_section = "Ecosystem Considerations", section_location = "after-discussion",
+#' @examples
+#'\dontrun{
+#' add_section(
+#'   new_section = "Ecosystem Considerations",
+#'   section_location = "after-discussion",
 #'   custom_sections = c("introduction.qmd", "model.qmd", "results.qmd", "discussion.qmd"),
```
Schiano-NOAA, line 19: This might need to be a list. Can you check this? @sbreitbart-NOAA

R/add_theme.R
```diff
@@ -1,14 +1,21 @@
...
 #' figure or table. Currently, the function is able to format objects from:
-#' ggplot (ggplot2), base r plot, flextable (flextable), gt tables (gt), and kable tables (kableExtra).
+#' ggplot (ggplot2), base r plot, flextable (flextable), gt tables (gt),
+#'  and kable tables (kableExtra).
 #' @export
 #'
-#' @examples add_theme(ggplot2::ggplot(data = cars, ggplot2::aes(x = speed, y = dist)) +
-#'   ggplot2::geom_point())
+#' @examples
+#'\dontrun{
```
Schiano-NOAA, line 13: in this situation, I feel like it would be OK to run, but probably doesn't matter

R/convert_output.R
```diff
@@ -1,15 +1,11 @@
 #' Convert Output
 #'
 #' Format stock assessment output files to a standardized format.
-#'
+
+#' @inheritParams create_template
```
Schiano-NOAA, line 5: In most of these functions, the arguments will be changes from the other PR so we don't really have to make many changes. This could cause issues when merging, but we'll figure it out

R/create_citation.R
```diff
@@ -7,12 +7,15 @@
 #' with `asar`.
 #' @export
 #'
-#' @examples create_citation(
-#'   title = "SA Report for Jellyfish",
+#' @examples
```
Schiano-NOAA, line 10: I think this might be another example that can be run if we want to bc all it does is create a string formed for cat. I would name the function something like `citation <- create_citation()` then end the example with cat(citation) so users can understand the use for the fxn

R/create_tables_doc.R
```diff
@@ -1,15 +1,22 @@
 #' Create Quarto Document of Tables
 #'
-#' @param resdir directory where the results file is located
-#' @param model_results name of the results file of assessment output
-#' @param model stock assessment model
-#' @param subdir subdirectory where the assessment report template is being stored
-#' @param include_all include all default tables for a stock assessment report
+#' @inheritParams create_template
```
Schiano-NOAA, line 3: ditto about the args in the other PR

R/create_template.R
```diff
@@ -104,7 +106,81 @@
...
+#'   type = "SAR",
+#'   prev_year = 2021,
+#'   custom = TRUE,
+#'   custom_sections = c("executive_summary", "introduction", "discussion"),
+#'   include_figures = TRUE,
+#'   include_tables = TRUE,
+#'   add_image = TRUE,
+#'   spp_image = "dir/containing/spp_image"
+#' )
+#' }
```
Schiano-NOAA, line 183: I'm not sure if this is the standard format, but I have seem examples that also end with the comment of don't run so people don't forget lol

R/create_title.R
```diff
@@ -5,10 +5,18 @@
 #' @return NULL
 #' @export
 #'
-#' @examples create_title(
-#'   office = "SEFSC", species = "Red Snapper", region = "South Atlantic",
-#'   year = 2024, type = "SAR", spp_latin = "Lutjanus campechanus"
+#' @examples
```
Schiano-NOAA, line 8: another situation like `create_citation` where the example could actually run, you can name the fxn, then cat(name)

Schiano-NOAA (changes requested):
This looks good overall. Please see my comments about some functions and other documentation. Thanks for doing this!

R/add_child.R
sbreitbart-NOAA: Yes, I believe so, since I didn't change the function's arguments at all- only separated them onto separate lines. In `create_template.R`, `add_child()` is used in a way that also suggests that x contains .qmd and the label doesn't: https://github.com/nmfs-ost/asar/blob/a46b0a4d50545fee673444152a1edfb1dfebe6b9/R/create_template.R#L512C1-L542C10

R/add_section.R
sbreitbart-NOAA: Running the examples with `custom_sections = c("introduction.qmd"...),` and `custom_sections = list("introduction.qmd"...` both seem to provide the same output in the SAR_species_skeleton.qmd. Would you prefer I use one option over the other? Or even use one format in one example, and the other format in another example?

R/add_theme.R
sbreitbart-NOAA: Should I remove the `\dontrun{}`?

R/convert_output.R
sbreitbart-NOAA: 👍 True. I'll keep this as-is, then, and fix it later.

R/create_citation.R
sbreitbart-NOAA: 👍 Edited as per your suggestion.

R/create_title.R
sbreitbart-NOAA: 👍 Edited as per your suggestion.

R/add_theme.R
Schiano-NOAA: We can keep for now

PR #65, opened 2024-10-25
Pkgdown vignettes

sbreitbart-NOAA:
# What is the feature?
* A new vignette about increasing the accessibility of the user's report

* Creating a vignette

* Don't think so


Schiano-NOAA (approved):
I'm a bit confused why the edits I made to the create_template fxn and README were added into here, but oh well. Let me figure out the merge conflicts then it will be OK

PR #68, opened 2024-10-29
build: added alt text to cover page's US Dept of Commerce seal

sbreitbart-NOAA:
# What is the feature?
* Added alt text to cover page's US Dept of Commerce seal

* Writing a line of LaTeX

* No

### DIFF
--- a/inst/resources/formatting_files/_titlepage.tex
+++ b/inst/resources/formatting_files/_titlepage.tex
@@ -107,7 +107,7 @@ $endfor$
 \vspace{1\baselineskip}
 
 %%%%%% Tagline at bottom
-\includegraphics[width=2cm]{support_files/us_doc_logo.png}\newline
+\pdftooltip{\includegraphics[width=2cm]{support_files/us_doc_logo.png}}{Seal of the United States Department of Commerce}\newline
 U.S. Department of Commerce\newline
 National Oceanic and Atmospheric Administration\newline
 National Marine Fisheries Service\newline
--- a/inst/resources/formatting_files/_titlepage.tex
+++ b/inst/resources/formatting_files/_titlepage.tex
@@ -107,7 +107,7 @@ $endfor$
 \vspace{1\baselineskip}
 
 %%%%%% Tagline at bottom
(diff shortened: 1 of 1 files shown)


inst/resources/formatting_files/_titlepage.tex
```diff
@@ -107,7 +107,7 @@
 \vspace{1\baselineskip}
 
 %%%%%% Tagline at bottom
-\includegraphics[width=2cm]{support_files/us_doc_logo.png}\newline
+\pdftooltip{\includegraphics[width=2cm]{support_files/us_doc_logo.png}}{Seal of the United States Department of Commerce}\newline
```
kellijohnson-NOAA, line 110: I am not sure why you are using pdftooltip here when you can just add the alternative text to the call to includegraphics. Also, the seal is an artifact rather than an actual figure in my opinion and you can tag it as such so that no alternative text is needed. From [W3](https://www.w3.org/TR/WCAG20-TECHS/PDF4.html) artifacts are "generally graphics objects or other markings that are not part of the authored content. Examples of artifacts include page header or footer information, lines or other graphics separating sections of the page, or decorative images."

sbreitbart-NOAA: Hi Kelli- thanks for your review. My responses:

_I am not sure why you are using pdftooltip here when you can just add the alternative text to the call to includegraphics._
- This mistake was because I'm so new to LaTeX and was trying to implement a solution I saw on StackOverflow. Thanks for sharing the better way to do that.

_Also, the seal is an artifact rather than an actual figure in my opinion and you can tag it as such so that no alternative text is needed. From [W3](https://www.w3.org/TR/WCAG20-TECHS/PDF4.html) artifacts are "generally graphics objects or other markings that are not part of the authored content. Examples of artifacts include page header or footer information, lines or other graphics separating sections of the page, or decorative images."_
- Thanks for letting me know. I agree that the seal is, in fact, in artifact. My next commit will include the .tex file with the alt text removed and the artifact tag added.

Schiano-NOAA (approved):

PR #88, opened 2024-11-26
Add minor change to convert_output arg defaults and format latin name in title

Schiano-NOAA:
# What is the feature?
*Set the output directory and save directory in the `convert_output` function to the working directory
*Estimate, uncertainty, and initial columns in the output data frame are now labelled as numeric
*Make latin species name in title italicized 

*change arguments
*add as.numeric to output df for respective columns
*add latex formatting fxn to template

*no

### DIFF
--- a/R/create_title.R
+++ b/R/create_title.R
@@ -21,6 +21,9 @@ create_title <- function(
     year <- format(Sys.Date(), "%Y")
   }
 
+  # Species latin name with italics latex fxn
+  spp_latin <- paste("\\textit{", spp_latin, "}", sep = "")
+
   # Create title dependent on regional language
   if (office == "AFSC") {
     if (is.null(complex)) {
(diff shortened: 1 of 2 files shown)


R/create_title.R
```diff
@@ -21,6 +21,9 @@ create_title <- function(
     year <- format(Sys.Date(), "%Y")
   }
 
+  # Species latin name with italics latex fxn
+  spp_latin <- paste("\\textit{", spp_latin, "}", sep = "")
```
sbreitbart-NOAA, line 25: The final string, in latex, should have one backslash. Does adding two here have a purpose for, e.g., escaping recognition in R?

sbreitbart-NOAA (commented):
Only one comment about the Latex. Otherwise looks good!

R/create_title.R
Schiano-NOAA: Yes, its necessary. When using cat() to form the tex file, the double slash is needed otherwise it messes up the formatting :/

PR #91, opened 2024-12-05
Integrate rda-based architecture

sbreitbart-NOAA:
# What is the feature?
* Integrating satf's rda-based figure/table/caption/alt text-generating architecture into asar

* Updating create_figures_doc.R, create_tables_doc.R, create_template.R to be able to intake rdas already generated with satf, or generate rdas and intake those rdas, to produce figures and tables with alt text and captions

* Yes- this relies on satf and impacts multiple scripts in the project


R/create_template.R
```diff
@@ -340,14 +402,38 @@ create_template <- function(
         }
       }
 
+      # run satf::exp_all_figs_tables() if rda files not premade
```
Schiano-NOAA, line 405: needs check for when users run create_template with no argument - this will probably break if they do so (maybe it doesn't though and you've addressed this). 

Maybe good to also add a switch like with convert_output to run this or not but set it TRUE by default

Schiano-NOAA (changes requested):
Looks good with the exception of one thing I noted. We just need to make sure we can run create_template with no args for users if that's what they want (while it's not suggested).

Also I noticed you pulled in some changes that I made/updated from main or dev. I have also been encountering this and having to do it over and over and it's annoying. Idk why the rebase won't fix this

Schiano-NOAA:
@sbreitbart-NOAA Do you know why the checks are failing?

sbreitbart-NOAA:
> @sbreitbart-NOAA Do you know why the checks are failing? Only the ubuntu is failing for other branches and I am working on that.

It's (at least partly, and maybe totally) because the tests haven't been updated to include instances of create_template() with the new parameters, which throw errors. Specifically, those errors arise because (back to your prior comment), create_template() breaks if rda_dir isn't specified (and therefore = FALSE) _and_ there aren't arguments provided to make rdas. I'll modify create_template() to work without any args and we'll see if that solves the failing tests.

R/create_template.R
sbreitbart-NOAA: Yes, definitely make this workable when create_template has no args. Great catch. Will add now

sbreitbart-NOAA:
> Looks good with the exception of one thing I noted. We just need to make sure we can run create_template with no args for users if that's what they want (while it's not suggested).
> 
> Also I noticed you pulled in some changes that I made/updated from main or dev. I have also been encountering this and having to do it over and over and it's annoying. Idk why the rebase won't fix this

I don't know the reason for the issue with having to rebase the same changes over and over again ☹️

Schiano-NOAA (approved):
Looks good. Have you tested create_template with no args to make sure the new code works?

sbreitbart-NOAA:
> Looks good. Have you tested create_template with no args to make sure the new code works?

Yup! In that case, no rdas are made and the resulting figures and tables .qmds are nearly blank with a warning message.

Schiano-NOAA:
> > Looks good. Have you tested create_template with no args to make sure the new code works?
> 
> Yup! In that case, no rdas are made and the resulting figures and tables .qmds are nearly blank with a warning message.

Perfect 👍 good to merge!

Issue #93, opened 2024-12-06
[Bug]: vignette not display as intended


Schiano-NOAA:
### Describe the bug

The following code chunk in the vignette will not be displayed as intended to the user:

https://github.com/nmfs-ost/asar/blob/a9b19fec03254fa533b7408c93f7b0078835adf8/vignettes/accessibility_guide.Rmd#L44-L46

### Which OS are you seeing the problem on?

Windows

### Which version of asar are you seeing the problem on?

dev branch

sbreitbart-NOAA:
PR to fix bug: https://github.com/nmfs-ost/asar/pull/95

sbreitbart-NOAA:
Closed as completed.

PR #95, opened 2024-12-06
fixed bug in a11y vignette

sbreitbart-NOAA:
# What is the feature?
* Fixing bug in https://github.com/nmfs-ost/asar/issues/93#event-15562576631

* Adding a "verbatim" code chunk around faulty chunk

* No


Schiano-NOAA (approved):
Solid 👍

Issue #62, opened 2024-10-18
[Bug]: Issues loading fish images


sbreitbart-NOAA:
### Describe the bug

The fish cover image won't load using absolute filepaths if using the add_image = TRUE and spp_image arguments. The image will only load once it's been placed in the directory with the new skeleton .qmd file, and spp_image calls the filename only- no filepath (e.g., "Rex_sole.png").

### Describe how to reproduce the bug

# This does work:
```
asar::create_template(
        new_template = TRUE,
        format = "pdf",
        office = "NWFSC",
        species = "Rex sole",
        spp_latin = "Glyptocephalus zachirus",
        year = 2010,
        add_image = TRUE, 
        spp_image = "Rex_sole.png"
    )
```

# This doesn't work:
```
asar::create_template(
        new_template = TRUE,
        format = "pdf",
        office = "NWFSC",
        species = "Rex sole",
        spp_latin = "Glyptocephalus zachirus",
        year = 2010,
        add_image = TRUE, 
        spp_image = "C:/Users/sophie.breitbart/Documents/R_projects/asar/inst/resources/spp_img/Rex_sole.png"
    )
```

### Which OS are you seeing the problem on?

Windows

### Which version of asar are you seeing the problem on?

0.0.0.9000

Schiano-NOAA:
This is failing because the addition of a species image should be just the image. For example "species_images.png". The user should move the image to the support_files folder where the current assessment is being rendered. I think it will be better to add the image to the folder rather than call it so then they can keep everything together and it will be there if they use asar in the future to make the next assessment.

sbreitbart-NOAA:
The support_files folder is only created after entering `create_template()`, right? It might be confusing for the user to have to use that function, add the image, and then potentially enter `create_template()` again so that R can find the image. Is that the proposed workflow, or am I missing something?

Schiano-NOAA:
Resolved with PR #104

PR #104, opened 2024-12-12
fix(cover image)

Schiano-NOAA:
* Standardizes the size of the cover image
* adjusts spp_image argument so users can add a path to the image and it uses it properly #62

# What is the feature?
*add spp image that is outside of the template

*copy image file to working folder and extract image file name

*no


sbreitbart-NOAA (approved):
Looks great! I think this will be easier for the user to understand.

Issue #56, opened 2024-10-07
Update create_citation


Schiano-NOAA:
Need to review different requirements for citations for each council and if the current setup in the function conflicts with these requirements.

- [ ] NPMF
- [ ] PFMC
- [ ] WPFMC
- [ ] NEFMC
- [ ] MAFMC
- [ ] SAFMC
- [ ] GMFMC
- [ ] CFMC

PR #107, opened 2024-12-17
added first draft of new FSC-specific citations

sbreitbart-NOAA:
# What is the feature?
* adding ability to create FSC-specific citations, as per issue: https://github.com/nmfs-ost/asar/issues/56

* expanding if/else chunks that make office-specific citations

* No

### DIFF
--- a/R/create_citation.R
+++ b/R/create_citation.R
@@ -121,7 +121,62 @@ create_citation <- function(
       title, ". ", off_title, ", ",
       "CITY, STATE. "
     )
-  } else if (office != "SEFSC") {
+
+    } else if (office == "AFSC") {
+      cit <- paste0(
+        "{{< pagebreak >}} \n",
+        "\n",
+        "Please cite this publication as: \n",
+        "\n",
+        author_list, ". ", year, ". ",
+        title,
+        ". North Pacific Fishery Management Council, Anchorage, AK. Available from ",
+        "https://www.npfmc.org/library/safe-reports/"
+      )
+
+    } else if (office == "NWFSC") {
+      cit <- paste0(
+        "{{< pagebreak >}} \n",
+        "\n",
+        "Please cite this publication as: \n",
+        "\n",
+        author_list, ". ", title, ".", year,
+         ". Prepared by ... . XX p."
+      )
+
+    } else if (office == "PIFSC") {
+      cit <- paste0(
+        "{{< pagebreak >}} \n",
+        "\n",
+        "Please cite this publication as: \n",
+        "\n",
+        author_list, ". ", year, ". ",
+        title, ". NOAA Tech. Memo. [TECH MEMO NUMBER]",
+        ", ", "XX p."
+      )
+
+  } else if (office == "SEFSC") {
+    cit <- paste0(
+      "'{{< pagebreak >}}' \n",
+      "\n",
+      "Please cite this publication as: \n",
+      "\n",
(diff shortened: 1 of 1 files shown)


R/create_citation.R
```diff
@@ -117,30 +117,76 @@ create_citation <- function(
...
+      )
+
+    } else if (office == "NWFSC") {
+      cit <- paste0(
+        "{{< pagebreak >}} \n",
+        "\n",
+        "Please cite this publication as: \n",
+        "\n",
+        author_list, ". ", title, ".", year,
+         ". Prepared by [COMMITTEE]. [XX] p."
```
Schiano-NOAA, line 144: I wonder if this can be automated...

R/create_citation.R
```diff
@@ -117,30 +117,76 @@ create_citation <- function(
...
+
+    } else if (office == "PIFSC") {
+      cit <- paste0(
+        "{{< pagebreak >}} \n",
+        "\n",
+        "Please cite this publication as: \n",
+        "\n",
+        author_list, ". ", year, ". ",
+        title, ". NOAA Tech. Memo. [TECH MEMO NUMBER]",
+        ", ", "[XX] p."
```
Schiano-NOAA, line 155: same as above

R/create_citation.R
```diff
@@ -117,30 +117,76 @@ create_citation <- function(
...
-      "{{< pagebreak >}} \n",
+      "'{{< pagebreak >}}' \n",
       "\n",
       "Please cite this publication as: \n",
       "\n",
-      author_list, ". ", year, ". ",
-      title, ". ", off_title, ", ",
-      loc_city, ", ", loc_state, ". "
+      "SEDAR. ", year, ". ", title, ". ",
+      "SEDAR, North Charleston SC. [XX] pp. ",
```
Schiano-NOAA, line 165: ditto

Schiano-NOAA (approved):
This looks good to me - I haven't test this. I am wondering if we can add code that would automate the number of pages (I feel like this should be possible)

sbreitbart-NOAA:
> This looks good to me - I haven't test this. I am wondering if we can add code that would automate the number of pages (I feel like this should be possible)

From a little googling, I'm not finding any quick fixes. I think it might be tricky. The workflow for calculating the page length might be: render to pdf with latex, then put the number back into the R script. I'm not sure R could accurately calculate page length before rendering. What if we added it as an issue for the future?

Schiano-NOAA:
> > This looks good to me - I haven't test this. I am wondering if we can add code that would automate the number of pages (I feel like this should be possible)
> 
> From a little googling, I'm not finding any quick fixes. I think it might be tricky. The workflow for calculating the page length might be: render to pdf with latex, then put the number back into the R script. I'm not sure R could accurately calculate page length before rendering. What if we added it as an issue for the future?

@sbreitbart-NOAA That works! Maybe just something to think about as an added feature, but if we can't do it, we can't do it 🤷

kellijohnson-NOAA:
You can use a call to latex within the template that puts in the number of page numbers on the second time that it is built.

sbreitbart-NOAA:
> You can use a call to latex within the template that puts in the number of page numbers on the second time that it is built.
> 
> ```
> \pageref{LastPage}{}
> ```

Awesome, thanks Kelli!! We put this into a new issue https://github.com/nmfs-ost/asar/issues/108 and will revisit after the first release.

PR #115, opened 2024-12-27
Integrate failsafe for rda files

Schiano-NOAA:
# What is the feature?
* Add a check so that if the satf::exp_all_figs_tables function doesn't work then the create_template function will still proceed, but it will just let the user know of the missing info

*Added try function around exp_all fxn

* no

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -422,6 +422,77 @@ create_template <- function(
         }
       }
 
+      # Convert output file if TRUE
+      if (convert_output) {
+        print("__________Converting output file__________")
+        if (tolower(model) == "bam" & is.null(fleet_names)) {
+          # warning("Fleet names not defined.")
+          convert_output(
+            output_file = model_results,
+            outdir = resdir,
+            file_save = TRUE,
+            model = model,
+            savedir = subdir,
+            save_name = paste(stringr::str_replace_all(species, " ", "_"), "_std_res_", year, sep = "")
+          )
+          # } else if (tolower(model) == "bam") {
+          #   convert_output(
+          #     output_file = model_results,
+          #     outdir = resdir,
+          #     file_save = TRUE,
+          #     model = model,
+          #     fleet_names = fleet_names,
+          #     savedir = subdir,
+          #     save_name = paste(sub(" ", "_", species), "_std_res_", year, sep = "")
+          #   )
+        } else {
+          convert_output(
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA:
@Schiano-NOAA can you please rebase with dev-1.0, then re-request review? The updated create_template's exp_all_figs_tables fxn has a few args that have been added over the past week.

Schiano-NOAA:
@sbreitbart-NOAA I was planning adding this directly into main with this PR, but regardless this should merge OK with the changes you made because the code just wraps around it. I can switch branches for the PR if you think that would be better?

Schiano-NOAA:
@sbreitbart-NOAA making additional changes to this as it now does not work completely as intended. Will ping you for review when ready

Schiano-NOAA:
@sbreitbart-NOAA please review

R/create_figures_doc.R
```diff
@@ -120,8 +125,12 @@ if (file.exists(file.path(rda_dir, 'spawning.biomass_figure.rda'))){\n
       ),
       "\n"
     )
+    } else {
+      message("Spawning biomass time series figure not created.")
+    }
 
     # B figure
+    if (any(grepl("biomass_figure.rda", list.files(file.path(rda_dir, "rda_files"))))) {
```
sbreitbart-NOAA, line 133: It's creating chunks to plot the biomass figure even if I don't have the biomass rda. Guessing it's related to the grepl fxn looking for biomass and finding spawning biomass?

R/create_template.R
```diff
@@ -529,7 +548,7 @@ create_template <- function(
           } else {
             create_tables_doc(
               subdir = subdir,
-              rda_dir = subdir
+              rda_dir = rda_dir
```
sbreitbart-NOAA, line 551: Good catch!

sbreitbart-NOAA:
Besides the one grepl error, looks good!

R/create_figures_doc.R
Schiano-NOAA, line 133:
```suggestion
    if (any(grepl("^biomass_figure.rda", list.files(file.path(rda_dir, "rda_files"))))) {
```

Schiano-NOAA: @sbreitbart-NOAA I added a symbol in the regrex that should fix that! Thanks for the catch

PR #139, opened 2025-01-13
Remove nmfspallete from dependencies temporarily

Schiano-NOAA:
# What is the feature?
* Removing dependency from package to allow download of package
* Comment out use of `nmfspallete` in add_theme function and its tests

*Remove `nmfspallete` from dependencies in DESCRIPTION

*yes. this will now allow users to properly plot with the nmfs colors; however, we have also temporarily removed instances of nmfspallete from `satf`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,6 @@ Imports:
     glue,
     gridExtra,
     naniar,
-    nmfspalette,
     officer,
     openxlsx,
     prodlim,
@@ -52,8 +51,6 @@ Suggests:
 VignetteBuilder: 
     knitr
 Remotes:
-    nmfs-fish-tools/nmfspalette,
-    nmfs-fish-tools/nmfspalette,
     nmfs-ost/satf,
     r4ss/r4ss
 Config/testthat/edition: 3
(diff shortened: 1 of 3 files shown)


sbreitbart-NOAA (commented):
nmfspalette is also in asar::add_theme and test-add_theme. But I believe that fxn and test can be removed since they're duplicated in satf (the better home for this fxn). So I'd either remove those files (and their documentation), or comment them out in those places as well.

Schiano-NOAA:
@sbreitbart-NOAA I noticed that as soon as the actions failed lol, but just noticing as well that we moved that function into satf. I think for v1.1 we can remove add theme and migrate the tests over to satf.

sbreitbart-NOAA:
> @sbreitbart-NOAA I noticed that as soon as the actions failed lol, but just noticing as well that we moved that function into satf. I think for v1.1 we can remove add theme and migrate the tests over to satf.

Sounds good. I take it you'll remove add_theme/migrate the tests to satf?

Schiano-NOAA:
> > @sbreitbart-NOAA I noticed that as soon as the actions failed lol, but just noticing as well that we moved that function into satf. I think for v1.1 we can remove add theme and migrate the tests over to satf.
> 
> Sounds good. I take it you'll remove add_theme/migrate the tests to satf?

Yeah I can do that once the package is fixed

PR #149, opened 2025-01-17
fix to incorrect reference in .bib file

iantaylor-NOAA:
# What is the feature?
* Fixes an incorrect reference in asar_references.bib. Sorry if this PR is to the wrong branch or you prefer a different process to edit the .bib file.

* I copied info from source URL which @okenk and @chantelwetzel-noaa tracked down. This matters more than you would expect because the document is a key data source that we have not consistently included in production assessments. Having an accurate reference will improve access to the report and make it more likely that all catch is accounted for.

* Definitely not.

### DIFF
--- a/inst/resources/asar_references.bib
+++ b/inst/resources/asar_references.bib
@@ -742,8 +742,9 @@
 }
 
 @techreport{rogers_species_2003,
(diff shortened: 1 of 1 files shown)


Schiano-NOAA:
@iantaylor-NOAA thanks! This is a great catch. This .bib file was added so the report renders properly when people test it out/use it initially. We are hoping to expand upon it in the future with PRs like this to increase its coverage, but it was not high priority enough to do it v1.0 or soon. 

We will go through the rest of the bib file to check for any remaining errors if any. If you see any more, please let me know :)

Issue #144, opened 2025-01-16
[Bug]: create_template rda issue


sbreitbart-NOAA:
### Describe the bug

`create_template` doesn't create rdas as expected, but it can handle existing rdas.

### Which OS are you seeing the problem on?

Windows

PR #151, opened 2025-01-21
Fix(create_template)

sbreitbart-NOAA:
# The impetus
- https://github.com/nmfs-ost/asar/issues/144

# What is the feature?
- Fix `create_template` so that rda files can be added as input, if already created, OR generated if model results are included as an argument.
- Update documentation

* in `create_template`, changed the `exp_all_figs_tables()` rda_dir argument from subdir to rda_dir. This changes the workflow so that the rda_files folder is always placed in rda_dir, not in the report folder. The issue was that the captions_alt_text.csv file was always being created in rda_dir, but `exp_all_figs_tables()` within `create_template()` was breaking because captions_alt_text.csv wasn't ever within the "report" folder. I chose to keep the rda_files folder in the rda_dir, instead of the report folder, because sometimes the user will be providing their own rda_files folder, and I wanted to keep the workflow consistent among users that did and did not provide their own rda_files folder.
* altered `create_tables/figures_doc()` calls in `create_template` so that rda_dir is always rda_dir (updated to match the change that the rda_files folder is always placed in rda_dir, not in the report folder.)

* `create_tables_doc()` & `create_figures_doc()` calls in `create_template`

# Tested?
* Tested with:
  * an SS3 model result csv
  * BAM model result csv
  * empty `create_template()` call
  * a `create_template` call with convert_output = T and model_results = Report.sso

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -104,8 +104,8 @@
 #' figures, tables, alt text, and captions with `satf`, rda_dir represents
 #' the location of the folder containing these .rda files ("rda_files").
 #' Otherwise, if the user has not used `satf` to make those .rda files already,
-#' those files will be generated automatically and placed within the "report"
-#' folder within the `file_dir`. The "rda_files" folder would have been
+#' those files will be generated automatically and placed within an "rda_files"
+#' folder within rda_dir. The "rda_files" folder would have been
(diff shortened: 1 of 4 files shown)


Schiano-NOAA (commented):
Looks good to me. The change seems pretty straight forward

Issue #155, opened 2025-01-24
Fix render with large tables


sbreitbart-NOAA:
Tables width > page width
Tables spanning >1 page
Landscape tables >1 page
- sablefish + bsb good examples

sbreitbart-NOAA:
Ideas that have not worked:

- adding these ["landscape mode" braces](https://quarto.org/docs/authoring/article-layout.html#landscape-mode) around results chunk in skeleton, and/or specific R chunk in 08_tables.qmd
- rotating the flextable (no option for this; only rotating text, which doesn't update the placement of the headers)
-

sbreitbart-NOAA:
Ideas that have shown promise:
1. Adding the 3 latex packages in the skeleton yaml under the "header-includes" line, and wrapping a plot chunk (e.g., label: 'tbl-indices-plot') in these lines, [as explained in this post](https://github.com/quarto-dev/quarto-cli/discussions/680#discussioncomment-2588520):

\newpage
\blandscape
Landscape
```{r}
summary(cars)
```
\elandscape

This forces the R chunk to be plotted in a landscape-oriented page, and creates output like this: ![Image](https://github.com/user-attachments/assets/2ee7def2-da57-4e0f-8662-15b23200cadb)

2. Adding code like this to that same chunk in 08_tables.qmd (replacing the above `summary(cars)` line), so that if the table is wider than the rotated page, the text is shrunken so that the table can fit in the landscape page.

Example code:

```
# set page width in inches
page_width <- 7.5

# identify flextable width
table_width <- flextable::flextable_dim(indices_table)[["widths"]] |> as.numeric()

if (table_width > page_width){
  indices_table <- indices_table |>
    flextable::fit_to_width(max_width = page_width )

  indices_table
} else {
  indices_table
}
```

This creates output like this: 

![Image](https://github.com/user-attachments/assets/3ed6e694-1a9e-405d-878d-e4ae136b0ed1)

sbreitbart-NOAA:
> [#2](https://github.com/nmfs-ost/asar/issues/2) could be a good option. We would just need to code it into the tables. This however does not help the users if they are making custom ones. This is a great start!

Turns out, I just needed to update quarto to get the landscape option to work 🤦 but it's much easier to use. It might be possible to automate the insertion of the landscape braces around the chunk if a table is too wide for portrait- will try this next. But I think we've hit a milestone of identifying how to switch a table's orientation from portrait to landscape, at least!

[...]

PR #158, opened 2025-01-28
Render large tables

sbreitbart-NOAA:
# What is the feature?
* Rendering large tables properly, as per https://github.com/nmfs-ost/asar/issues/155

* Quantifying each table's width in `create_tables_doc`, then customizing the appropriate chunk in the 08_tables.qmd file so that tables wider than 5" are placed in a landscape-orientation page. Tables between 5-12" wide are forced to shrink in width to 7.5" (this shrinks the text to fit). Tables > 12" wide are split into multiple tables, saved as a new rda, and placed in individual chunks in the 08_tables.qmd doc.

* No


R/create_tables_doc.R
```diff
@@ -111,7 +111,7 @@ if (file.exists(file.path(rda_dir, 'bnc_table.rda'))){\n
           if (orient_landscape_bnc == TRUE){
             paste0(
               "bnc_table |>
-                flextable::fit_to_width(max_width = 7.5) |>
+                flextable::fit_to_width(max_width = 8.5) |>
```
Schiano-NOAA, line 117: does this only switch to landscape if its ?> page size?

Schiano-NOAA (commented):
Looks good. I have not tested but looks like in practice it should just check if the table needs to be rotated or not

Schiano-NOAA:
Next thing to think about is how this could be applied for custom tables? or do we just state a recommended practice to follow?

sbreitbart-NOAA:
> Looks good. I have not tested but looks like in practice it should just check if the table needs to be rotated or not

Thanks! What I hope is one of the last components of this PR, which I'm working on now, is designing a way to accommodate tables that need landscape orientation but are too wide to be shrunken with the existing workflow (e.g., red snapper, which has a 60" wide indices table without shrinking, and with shrinking, has text that is too small to read).

sbreitbart-NOAA:
> Next thing to think about is how this could be applied for custom tables? or do we just state a recommended practice to follow?

Great point. I'll write up an explanation for our workflow and explain how the user can do it manually for custom tables in the FAQs and manual.

Schiano-NOAA (changes requested):
I reran the create_tables_doc function from a previous example and it worked...however, it repeated the table caption over and over in the list of tables in the beginning of the doc and broke it up incorrectly in the document itself (I think). Guess we will need to chat about this next week. I can retest on monday with a fresh mind just in case :)

sbreitbart-NOAA:
> I reran the create_tables_doc function from a previous example and it worked...however, it repeated the table caption over and over in the list of tables in the beginning of the doc and broke it up incorrectly in the document itself (I think). Guess we will need to chat about this next week. I can retest on monday with a fresh mind just in case :)

Thanks for testing. Can you please explain how it's breaking the tables up incorrectly? And what would you prefer the table captions look like?

Schiano-NOAA:
> > I reran the create_tables_doc function from a previous example and it worked...however, it repeated the table caption over and over in the list of tables in the beginning of the doc and broke it up incorrectly in the document itself (I think). Guess we will need to chat about this next week. I can retest on monday with a fresh mind just in case :)
> 
> Thanks for testing. Can you please explain how it's breaking the tables up incorrectly? And what would you prefer the table captions look like?

image

Schiano-NOAA:
@sbreitbart-NOAA I think the tables are broken up fine but I wonder if it's possible to add into the caption the fleets that are included in the separated table

sbreitbart-NOAA:
@Schiano-NOAA I've updated the workflow so that when a table is split, the header values for each table are added in parentheses to the end of the table caption, like this:

image

Is this change sufficient to approve the PR? Or should I make more changes to the workflow?

Schiano-NOAA:
> @Schiano-NOAA I've updated the workflow so that when a table is split, the header values for each table are added in parentheses to the end of the table caption, like this:
> 
> image
> 
> Is this change sufficient to approve the PR? Or should I make more changes to the workflow?

This should be good. Let me do a quick test

PR #167, opened 2025-02-24
Make asar compatible with stockplotr

sbreitbart-NOAA:
# What is the feature?
* Updating asar's functions to work with stockplotr's, after 1) updates to exp_all_figs_tables and 2) name change from satf to stockplotr.

* Adding in the new args to create_template's exp_all_figs_tables, and removing old, and replacing satf with stockplotr

* Yes

### DIFF
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -83,7 +83,6 @@ roboto
 rowid
 RStudio
 saltatrix
-satf
 sd
 se
 snakecase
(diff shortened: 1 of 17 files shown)


inst/WORDLIST
Schiano-NOAA, line 1: Can you remove satf from the wordlist? This will also make sure the word gets pinged if it's used on accident

Schiano-NOAA (approved):

PR #173, opened 2025-03-07
Fix bug with references rendered in wrong location

Schiano-NOAA:
# What is the feature?
* References are placed in their proper section
* new citation in bibliography for {asar}
* acknowledgement for use of {asar} in respective section

**Intended to be placed into main as a hotfix rather than with a new feature**

* Uncomment placement notation in references.qmd
* added new citation to .bib file
* added new statement line to acknowledgements.qmd

*no

### DIFF
--- a/inst/templates/skeleton/06_acknowledgments.qmd
+++ b/inst/templates/skeleton/06_acknowledgments.qmd
@@ -4,4 +4,4 @@
 
 <!--- Please leave in the following: --->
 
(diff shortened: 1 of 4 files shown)


inst/templates/skeleton/06_acknowledgments.qmd
Schiano-NOAA, line 1: @sbreitbart-NOAA I would appreciate feedback on this statement. Should it be this long? Do you think the first sentence only is sufficient?

Thanks!

PR #174, opened 2025-03-20
Feature draft

Schiano-NOAA:
# What is the feature?
*Default skeletons contain draft watermark - current users can update their skeleton by running create_template(rerender_skeleton=TRUE)
* New function to remove the watermark called "remove_draft"

*Added notation in YAML, for pdf, and after YAML, for html, to produce a watermark indicating the assessment is in draft.
*the new function, remove_draft, takes away this notation in the skeleton

*no


Schiano-NOAA:
@sbreitbart-NOAA testing for this feature should only have to require running a blank create_template then rendering. I checked for both pdf and html bc there are differences between how each are implemented. Thanks!!

Schiano-NOAA:
> Great feature! Question: Why is the watermark different for pdf vs html? I'd recommend choosing one for both formats if there's not a strong reason for two watermarks. Suggestion: Make remove_draft.R into a full function, with argument explanations, examples, etc. so that the user can learn exactly how this function works. I had to realize that I had to set my wd to "report" or else this wouldn't work (only knew this from previous conversations about skeleton-render. I think adding those instructions to the fxn documentation is essential for the fxn to be used.

Sounds great! Thanks for the suggestions. I totally forgot to add the documentation to the function so great catch on that. I will make those changes then merge into skeleton-render branch

PR #175, opened 2025-03-26
update DESCRIPTION to move unncessary dependencies to suggests

Schiano-NOAA:
# What is the feature?
* Reduce dependencies

*Move packages that are not directly references in a fxn to suggests

*no

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,16 +24,13 @@ Imports:
     flextable,
     forstringr,
     fs,
-    ggplot2,
     glue,
     gridExtra,
     naniar,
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA:
Do you have ideas for how to test this? Ensuring no errors when calling the package, calling all functions, ...?

Schiano-NOAA:
> Do you have ideas for how to test this? Ensuring no errors when calling the package, calling all functions, ...?

These should be fine to remove from dependencies as they are not referenced anywhere as dependent packages in our functions. They were reported in a check so I moved them to suggests

PR #188, opened 2025-04-07
Move two utils fxns to individual R files

sbreitbart-NOAA:
# What is the feature?
* Moved ID_tbl_width_class() and export_split_tbls() from utils to own .R files
* Made unit tests for each fxn

* Moving to separate .R files

* No


R/ID_tbl_width_class.R
```diff
@@ -0,0 +1,50 @@
+#' Identify table width class
+#'
+#' @inheritParams render_lg_table
+#' @param portrait_pg_width The amount of space between the margins of a
+#' portrait-oriented page, in inches. Represents the threshold for the maximum
+#' width of a table that can be rendered on a portrait page before it needs to
+#' be resized, rotated, and/or split across multiple pages.
+#'
+#' @return The width class of a table: regular, wide, or extra-wide. The result
```
Schiano-NOAA, line 9: Can you provide additional information on what the output would look like? Like it says it "will determine whether the table...", but is this the output? Like a TRUE/FALSE or something else?

Schiano-NOAA (approved):
Looks good to me

PR #211, opened 2025-04-24
Hotfix: custom section option

Schiano-NOAA:
# What is the feature?
* Fixed issue when custom = TRUE where there was duplicated code

* Removed the duplicated code
* Added conditions to add in the figures and table docs
* added new feature for rerender that will only add new authors to the documents

* yes,. other branches


Schiano-NOAA:
> This does fix the repeated code issue, but it won't render because the tables and figures child chunks still have the same label ('NA') and need to be differentiated

This is interesting because I have a check in there that they should be added. I will debug and ping you after.

Schiano-NOAA:
> > This does fix the repeated code issue, but it won't render because the tables and figures child chunks still have the same label ('NA') and need to be differentiated
> 
> This is interesting because I have a check in there that they should be added. I will debug and ping you after.

@sbreitbart-NOAA I just ran this on my machine and the tables and figures chunks have correct labels. What code are you running (so I can reproduce it)?

sbreitbart-NOAA:
> > > This does fix the repeated code issue, but it won't render because the tables and figures child chunks still have the same label ('NA') and need to be differentiated
> > 
> > 
> > This is interesting because I have a check in there that they should be added. I will debug and ping you after.
> 
> @sbreitbart-NOAA I just ran this on my machine and the tables and figures chunks have correct labels. What code are you running (so I can reproduce it)?

I'm running `create_template(custom = TRUE, custom_sections = c("executive_summary"))`

sbreitbart-NOAA:
> > > > This does fix the repeated code issue, but it won't render because the tables and figures child chunks still have the same label ('NA') and need to be differentiated
> > > 
> > > 
> > > This is interesting because I have a check in there that they should be added. I will debug and ping you after.
> > 
> > 
> > @sbreitbart-NOAA I just ran this on my machine and the tables and figures chunks have correct labels. What code are you running (so I can reproduce it)?
> 
> I'm running `create_template(custom = TRUE, custom_sections = c("executive_summary"))`
@Schiano-NOAA Just realized there are two branches- hotfix-custom-sections and hotfix-custom-sec. I was on the former but this PR is for the latter. I'll try that one now

sbreitbart-NOAA (approved):

Schiano-NOAA:
@sbreitbart-NOAA when you have the time, could you check that when you run custom=T and add a section where the section location is in another section?

Here some test code:
```
create_template(
    custom = TRUE,
    add_section = "Management",
    section_location = "in-data"
)
```

PR #215, opened 2025-04-29
Fix issue with add_alttext

Schiano-NOAA:
# What is the feature?
* fix(add_alttext): only add alttext for figures in loop for fxn

* reduced objects that are run through the loop

* no

### DIFF
--- a/R/add_alttext.r
+++ b/R/add_alttext.r
@@ -164,8 +164,9 @@ add_alttext <- function(
   obj_files <- list.files(file.path(rda_dir, "rda_files"))
 
   # read all files in obj_files and put into list
+  figures_list <- grep("figure", obj_files)
   alt_text_list <- list()
-  for (i in seq_along(obj_files)) {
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA (commented):
This works in that I can add alt text to the .tex file and finish compiling, but I don't see the alt text in the PDF

Schiano-NOAA:
> This works in that I can add alt text to the .tex file and finish compiling, but I don't see the alt text in the PDF

The way it's set up, you have to have add_tagging run with it as well so add_accessibility is the final/function to run to add this in.

sbreitbart-NOAA:
> > This works in that I can add alt text to the .tex file and finish compiling, but I don't see the alt text in the PDF
> 
> The way it's set up, you have to have add_tagging run with it as well so add_accessibility is the final/function to run to add this in.

Gotcha. Then this looks good to me!

PR #216, opened 2025-04-29
Change default text

Schiano-NOAA:
# What is the feature?
* making change to default asar template to render on linux

* change default font
* remove alt={} notation in DOC image

* no


inst/WORDLIST
```diff
@@ -60,7 +60,6 @@ gp
 img
 is’
 knitr
-kobe
```
sbreitbart-NOAA, line 63: Could you please rebase? These words shouldn't be removed from the wordlist

Schiano-NOAA: This is what happened after the rebase. They must have been removed in dev then

Schiano-NOAA: I have no idea why these are getting removed. I rebased to dev

sbreitbart-NOAA: Hmm. Maybe add them back in manually, then? Because I see them in the dev wordlist

sbreitbart-NOAA (commented):
I think you're almost there! I noticed that the figure captions are still in a serif font, and the table fonts look different too (is it just Arial? if so, then what about changing the report's fonts from arial narrow to arial to match the tables?). I think Arial would be easier to read, if it's in line with general branding guidance (which I'm sure you're aiming to be consistent with).

Schiano-NOAA:
@sbreitbart-NOAA I'm not sure which font we should go with so I am fine with switching to arial. I am not entirely sure which fonts do or don't work with linux. I looked at the other font and I think its basically the same as default quarto (which I really don't want)

sbreitbart-NOAA:
> @sbreitbart-NOAA I'm not sure which font we should go with so I am fine with switching to arial. I am not entirely sure which fonts do or don't work with linux. I looked at the other font and I think its basically the same as default quarto (which I really don't want)

Oh, my bad- Eli said this a few weeks ago:
"Arial is Microsoft font so not generically available on Linux and DejaVu Sans (getting injected somewhere because you only see in the generated tex). Latin Modern Sans will be in any TexLive installation."
[Here is a post](https://tex.stackexchange.com/questions/59403/what-font-packages-are-installed-in-tex-live) explaining which fonts are available by default in TexLive.

sbreitbart-NOAA (changes requested):
This renders locally, but Jupyterhub isn't rendering the report because 

ERROR: 
compilation failed- error
Package fontspec Error: 
(fontspec)                The font "Arial" cannot be found; this may be but
(fontspec)                usually is not a fontspec bug. Either there is a
(fontspec)                typo in the font name/file, the font is not
(fontspec)                installed (correctly), or there is a bug in the
(fontspec)                underlying font loading engine (XeTeX/luaotfload).

I think we should use a font that's on the list of fonts installed with TexLive.

sbreitbart-NOAA:
It works on jupyterhub! 🎉

PR #218, opened 2025-05-01
Glossary notation

Schiano-NOAA:
# What is the feature?
* Add in-header and after-body notation for glossary

* add notation adjustment to format_quarto and create_inheader_tex

* no

**Note: this will not be ready until we upload a new glossary in this PR**


Schiano-NOAA:
@sbreitbart-NOAA not sure if we want to wait to merge this in until after the glossary. Is the line for copying in the report_glossary.tex file in the other PR with the acronyms?

inst/WORDLIST
```diff
@@ -60,7 +60,6 @@ gp
 img
 is’
 knitr
-kobe
```
sbreitbart-NOAA, line 63: please add these words back

sbreitbart-NOAA:
I think we should merge the add-acronyms branch into this one, and then merge this one into dev. I'll have to rebase add-acronyms but otherwise the report_glossary.tex file is present and situated within the expected workflow where create_template() places it in the report folder

PR #221, opened 2025-05-02
Fix issue page break rendering before citation

Schiano-NOAA:
# What is the feature?
* see title
* hotfix(citation): fix bad notation before pagebreak causing the notation to render as text

* change glue to paste

* no

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -1057,7 +1057,7 @@ create_template <- function(
       if (rerender_skeleton) {
         if (!is.null(title) | !is.null(species) | !is.null(year) | !is.null(author)) {
           citation_line <- grep("Please cite this publication as:", prev_skeleton) + 2
-          citation <- glue::glue("{{< pagebreak >}} \n\n Please cite this publication as: \n\n {prev_skeleton[citation_line]}\n\n")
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA (commented):
This works/looks good to me, but I'm not sure what the issue was originally (having trouble reproducing it)

Schiano-NOAA:
> This works/looks good to me, but I'm not sure what the issue was originally (having trouble reproducing it)

When you rerendered an existing skeleton, the {{ <pagebreak> }} notation only included one curly bracket which doesnt make a page break on render

kellijohnson-NOAA:
FYI you just need to double the brackets in glue to make it print two sets, i.e.,
```
glue::glue("{{happy}} vs {{{{happy}}}}")
# Results in
{happy} vs {{happy}}
```

Schiano-NOAA:
> FYI you just need to double the brackets in glue to make it print two sets, i.e.,
> 
> ```
> glue::glue("{{happy}} vs {{{{happy}}}}")
> # Results in
> {happy} vs {{happy}}
> ```

Ah great to know! Kind of comincal with the double set. I was initially trying to just add a \ to see if it would ignore it like regex, but didn't realize you could add the extra brackets

Issue #223, opened 2025-05-02
[Bug]: city/state not added into affiliations


Schiano-NOAA:
### Describe the bug

The affiliation ends at the address for authors

### Describe how to reproduce the bug

add any author to a template

okenk:
The trick is editing [this line of the title page template](https://github.com/nmfs-ost/asar/blob/main/inst/resources/formatting_files/_titlepage.tex#L99)

I tried
`{$it.address$, $it.city$, $it.state$, $it.postal-code$}%`

And it prints:
Address, City, , Zip

In other words, "state" is not the correct key word to refer to the state variable from the YAML. I solved this temporarily by removing "state" from the YAML and writing the city variable in the format "City, State". Then the line of code for the title page is:
`{$it.address$, $it.city$, $it.postal-code$}%`

PR #226, opened 2025-05-06
Fix issue with create_citation and multiple matches to last name

Schiano-NOAA:
# What is the feature?
* Fixed issue where when there is more than one author last name match in the data base, the code breaks

* Fixed the error in the code
* implemented a fail safe so the author is still added

* no

### DIFF
--- a/R/create_citation.R
+++ b/R/create_citation.R
@@ -41,7 +41,7 @@ create_citation <- function(
       # Check and fix if there is more than one author with the same last name
       if (nrow(primauth_loc) > 1) {
         primauth_loc <- utils::read.csv(system.file("resources", "authorship.csv", package = "asar", mustWork = TRUE)) |>
-          dplyr::filter(last == unlist(strsplit(author[1], " "))[1])
+          dplyr::filter(first == unlist(strsplit(author[1], " "))[1])
       }
 
       office_loc <- utils::read.csv(system.file("resources", "affiliation_info.csv", package = "asar", mustWork = TRUE)) |>


sbreitbart-NOAA:
Just began testing. Using the following code doesn't produce a skeleton:

`create_template(author = "John Snow")`

sbreitbart-NOAA (commented):
Nice job- this is definitely improved! The only thing that's still not working is that the information after the author's road doesn't show up- like the state and zip code. Like this:

1. NOAA Fisheries Alaska Fisheries Science Center, 17109 Point Lena Loop Road
2. NOAA Fisheries Northeast Fisheries Science Center, 166 Water Street
3. NOAA Fisheries Northwest Fisheries Science Center, 2725 Montlake Boulevard East

kellijohnson-NOAA:
> Nice job- this is definitely improved! The only thing that's still not working is that the information after the author's road doesn't show up- like the state and zip code.

Are you referring to the suggested citation below this text "Please cite this publication as:" or in the actual affiliation at the top of the title page?

sbreitbart-NOAA:
> > Nice job- this is definitely improved! The only thing that's still not working is that the information after the author's road doesn't show up- like the state and zip code.
> 
> Are you referring to the suggested citation below this text "Please cite this publication as:" or in the actual affiliation at the top of the title page?

The actual affiliation at the top of the page. I know Sam was alerted about that [issue](https://github.com/nmfs-ost/asar/issues/223) and I thought was working to integrate it into this PR, but maybe not? If not, this looks good to me 👍

kellijohnson-NOAA:
It is not, that happens with the yaml header. I think that there is overlapping code here that could be alleviated with some helper functions but not sure if we want to do that now or in a different PR. Let me know if you want help with that.

Schiano-NOAA:
@sbreitbart-NOAA @kellijohnson-NOAA yep! Kelli is right. I was thinking about addressing that in the PR, but since it's not fatal to the function, I'm going to do it in a different PR into dev

Schiano-NOAA:
> It is not, that happens with the yaml header. I think that there is overlapping code here that could be alleviated with some helper functions but not sure if we want to do that now or in a different PR. Let me know if you want help with that.

Will definitely follow up with you for this. Probably will flag you to contribute or for review when I get to it.

PR #231, opened 2025-05-07
fix(create_citation): Fixes too many . in citation

kellijohnson-NOAA:
# What is the feature?
* Simplifies the code in `create_citation()` to use tidyverse style principles. There is one major change in that the citations no longer have initials before the second and subsequent authors. Instead, all initials come after the last name.

* Removes several if{}else{} statements and vectorizing things.

* There are several comments left in the code to explain what I did. Those will need to be removed before merging. More tests could also be written but I wasn't sure if this is actually what you would want.

@Schiano-NOAA and @sbreitbart-NOAA I honestly didn't mean to revamp this function I just wanted to figure out why there were too many ".", e.g., "Oshima, M.." in the citation within the rendered html for the case study and I got carried away. I think that there might be another PR or branch that is also working on this function but I am unsure.


Schiano-NOAA:
@kellijohnson-NOAA Thanks for this! I was actually working on this yesterday and planned to finish it up. I also had the same circumstance where when I went to fix the issue, I ended up revamping the function. I will incorporate a mix of the updates I made yesterday and those in this PR for the fix (especially when it comes to improving the function)

It was one of those situations where you question what your past self was thinking lol

kellijohnson-NOAA:
Sounds good @Schiano-NOAA 😄. Feel free to just close this PR or do whatever you need to do with it. I am also happy to jump on a call and do some group coding if you would like.

Schiano-NOAA:
Contributions of this PR added into #226

Issue #187, opened 2025-04-07
[Feature]: change affiliation default for new author argument


Schiano-NOAA:
### Is your feature request related to a problem? Please describe.

Need to add an author that is not in the data base and can not render bc the affiliation is not properly formatted

### Describe the solution you would like.

set defaults for all affiliation entries in yaml
- name:
  affiliation:
     - name:
     - address:
     - city:
     - state:
     - postal code:

### Describe alternatives you have considered

NA

PR #234, opened 2025-05-08
Fix affiliation default for add_author argument

sbreitbart-NOAA:
# What is the feature?
* When `create_template(add_author = "")` is used, the skeleton .qmd's yaml will include lines that allow the user to enter their own affiliation info, as per https://github.com/nmfs-ost/asar/issues/187

* Editing `create_template`

* No


R/create_template.R
```diff
@@ -812,7 +812,12 @@ create_template <- function(
             if (is.na(auth$office)) {
               paste(
                 "  ", "- name: ", "'", auth$name, "'", "\n",
-                "  ", "  ", "affiliations: ", "NO AFFILIATION", "\n",
+                "  ", "  ", "affiliations:", "\n",
```
Schiano-NOAA, line 815: Can you add the following next to their respective entries:
```
name: '[organization]'
address: '[address]'
city: '[city]'
state: '[state]'
postal-code: '[postal code]'
```

Just so there is an entry and it will draw people's eye to fill out - unless the template renders without these and it's nbd. I could see both ways

Schiano-NOAA (approved):

R/create_template.R
sbreitbart-NOAA: The template does render without anything filled in, but I think it's a good idea to add these signifiers too. Will add in the next commit 👍

PR #235, opened 2025-05-08
add quarto option to code coverage workflows

k-doering-NOAA:
# What is the feature?
* Add an option to install quarto before calculating coverage, which is necessary since quarto is a system dependency that asar has.

* Specify to install quarto in the coverage github action workflows

* No

Note: I wasn't certain about when the workflows should run given the use of main and dev branches - on pushes to main or dev? PR's to main or dev? Happy to discuss this so we can figure out solutions!


k-doering-NOAA:
Those are possibilities, let's talk through scenarios to decide what makes sense in our meeting today (or, at least so we can get clear on potential options so the workflows team can discuss as a whole and decide later!)

k-doering-NOAA:
@k-doering-NOAA needs to get these commits onto the dev branch, then reopen the PR!

PR #236, opened 2025-05-09
Fix author affiliation in title page

Schiano-NOAA:
# What is the feature?
* Combines city and state in the title page so it renders - temporary fix since we will be altering authorship approach soon anyway

* new notation in _titlepage.tex and change to authorship portion in create_template

* no


sbreitbart-NOAA (approved):
Works for me!

PR #242, opened 2025-05-13
Glossary documentation

sbreitbart-NOAA:
# What is the feature?
* New documentation explaining how to use acronyms and the glossary

* Editing vignettes

* No

### DIFF
--- a/vignettes/accessibility_guide.Rmd
+++ b/vignettes/accessibility_guide.Rmd
@@ -187,6 +187,10 @@ When you use an acronym, it will be included in a glossary table that is automat
 
 The glossary ("report_glossary.tex") is located in your report folder.
 
+Remember that your changes will only be reflected in your report folder's glossary .tex file, not in the file stored in the `asar` package's inst/ folder. If you regenerate your report folder, that main file will overwrite your edited version unless you have indicated otherwise.
+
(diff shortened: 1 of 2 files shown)


vignettes/accessibility_guide.Rmd
```diff
@@ -181,11 +181,51 @@ Looking for more resources for writing alt text? Check out the [NOAA Library's w
...
+
+`\newacronym{<label>}{<acronym ("short form")>}{<meaning ("long form")>}`
+
+Here is an example for NOAA:
+
+`\newacronym{noaa}{NOAA}{National Oceanic and Atmospheric Administration}`
+
+#### How to edit a glossary entry
+
+You can edit any entry as needed.
```
Schiano-NOAA, line 210:
```suggestion
You can edit any entry as needed. Remember that this change will not be reflected in the file from asar.
```

Schiano-NOAA: Maybe there should be a way to contribute? People could always make a PR and add to the file that way

Schiano-NOAA (approved):

vignettes/accessibility_guide.Rmd
sbreitbart-NOAA: Good idea. I'll add that, and your previous suggestion (with a little more detail)

PR #244, opened 2025-05-15
Add back glossary

sbreitbart-NOAA:
# What is the feature?
* Add back files used to create glossary (from asar.glossary repo)
* Create GH action that will make PR when asar::export_glossary() changes (indicating that report_glossary.tex file will have changed too)

*

*


sbreitbart-NOAA:
@Schiano-NOAA I believe this is finally ready for review! I'm still working on the GH action- meeting with Kathryn again next week to hopefully troubleshoot it- but that only has a manual trigger, so it shouldn't run otherwise.

inst/glossary/acronym_metadata.txt
Schiano-NOAA, line 1: Now that I'm thinking about it. This might fit well as a markdown document instead of txt so it's more readable in GitHub

PR #247, opened 2025-05-20
Improvements to convert_output

Schiano-NOAA:
# What is the feature?
* Removes unnecessary arguments from convert_output
* removes user ability to convert output within create_template
* adjusts create_template and dependent functions so they work (cleanup-pkg does not currently work)

* see functions for changes

* yes - major change in design


R/convert_output.R
```diff
@@ -2,19 +2,10 @@
 #'
 #' Format stock assessment output files to a standardized format.
 #'
-#' @param output_file Assessment model output file (e.g., the
+#' @param output_file Assessment model output file path (e.g., the
```
sbreitbart-NOAA, line 5: This should be updated to reflect example paths instead of the previous example files

R/convert_output.R
```diff
@@ -25,34 +16,25 @@
...
-#'         There are 2 options for adding data to the function. (1) Add the full
-#'         path with the file name in output.file or (2) output.file is the file
-#'         name and outdir is the path to the file without a trailing forward slash.
 #'
 #' @export
 #'
 #' @examples
 #' \dontrun{
 #' convert_output(
 #' output_file = "Report.sso",
```
sbreitbart-NOAA, line 25: Same comment as above re: updating with an example path vs. file

R/create_template.R
```diff
@@ -207,48 +188,37 @@ create_template <- function(
     new_template = TRUE,
     format = c("pdf", "docx", "html", NA),
     office = c("AFSC", "PIFSC", "NEFSC", "NWFSC", "SEFSC", "SWFSC"),
-    species = NA,
-    spp_latin = NA,
-    year = NA,
-    file_dir = NA,
+    species = NULL,
```
sbreitbart-NOAA, line 191: many of these NULLs should be NA, right?

R/create_citation.R
```diff
@@ -14,12 +14,11 @@
 #' )
 #'
 create_citation <- function(
-    author = NULL,
+    author = "",
```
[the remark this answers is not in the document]
Schiano-NOAA: Empty string or whatever gets placed in the default which is typically an empty string or something like "SPECIES"

R/create_template.R
Schiano-NOAA: Yeah I think I selected the wrong thing when fixing merge conflicts

R/create_yaml.R
```diff
@@ -50,17 +50,17 @@ create_yaml <- function(
     author = NA,
     add_author = NA,
     add_image = FALSE,
-    spp_image = NA,
-    species = NA,
```
sbreitbart-NOAA, line 54: These should also be NA, right?

sbreitbart-NOAA: Edit: should these be empty strings, instead of NULL?

R/create_template.R
Schiano-NOAA: Wait no on review, theses should be NULL - I changed them back to NULL from NA because it was doing the same thing and saved me time changing the checks from is.null to is.na which is doing the same thing. These defaults will probably get changed during this entire cleanup-pkg process but not in this PR

Schiano-NOAA:
@sbreitbart-NOAA I have made the changes you requested. In your example, I cannot replicate the issue you are having with any other data sets. I tried with the one you mentioned in the example.

Also, I am a little confused because it sounds like you are having the issue for the Hake example but the screen shot df is named skate.

Either way, this might be an issue with the plot and not converter since landings have always given some issue and need revising. (aka does not yet incorporate era into filtering)

sbreitbart-NOAA (approved):

Schiano-NOAA:
@sbreitbart-NOAA with this merge, you should be good to work on create_template changes as needed

PR #250, opened 2025-05-20
Remove nmfspalette from asar

sbreitbart-NOAA:
# What is the feature?
* Adding BugReports link to DESCRIPTION
* Removing mentions of nmfspalette (@Schiano-NOAA we don't need those, do we? Because the function that applies nmfspalette theming (add_theme) is in stockplotr

*

*

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,7 +45,6 @@ Suggests:
     gt,
     kableExtra,
     knitr,
-    nmfspalette,
     parallel,
     quarto,
     r4ss,
@@ -57,7 +56,6 @@ Suggests:
 VignetteBuilder: 
     knitr
 Remotes:
-    nmfs-ost/nmfspalette,
     nmfs-ost/stockplotr,
     r4ss/r4ss
 Config/testthat/edition: 3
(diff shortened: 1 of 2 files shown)


sbreitbart-NOAA:
I'm delaying substituting nmfspalette with viridis in stockplotr for a bit because it's more complicated than I thought. I think we should save this for the future, after we've taken the steps to remove exp_all_figs_tables from create_template

Schiano-NOAA:
> I'm delaying substituting nmfspalette with viridis in stockplotr for a bit because it's more complicated than I thought. I think we should save this for the future, after we've taken the steps to remove exp_all_figs_tables from create_template

I think we are good to merge this in, yeah?

Schiano-NOAA:
@sbreitbart-NOAA but this is in asar? Isn't nmfspalette a dependency of stockplotr? Since we no longer run exp_all_figs_tabs in asar then it should be okay right?

Schiano-NOAA:
@sbreitbart-NOAA Awesome, I'll run checks locally and rebase then merge 😄

Issue #201, opened 2025-04-16
[Feature]: time stamp & readme


BenWilliams-NOAA:
### Is your feature request related to a problem? Please describe.

Chantel mentioned in an issue
> ... If the consensus is that everything should be saved in the report file, I suggest adding something in the written files where a user can easily verify what model was used to create these files.

I think that is a great idea in general and could be added as both a message and the report can produce a README.md with associated information a la [`accepted_model()`](https://github.com/afsc-assessments/afscdata/blob/b3145f2082fa845bb7fa0642c68bde0f7679ac4c/R/utils.R#L203) which either creates or appends a description of which file was used/approved the previous year. 

### Describe the solution you would like.

Something like this message (pointed to wherever makes sense) from `asar::create_template()` would be great.

```
message(paste("Report is based upon model output from", model_results, 
              "\nthat was last modified on:", 
              file.info(here::here(resdir, model_results), extra_cols = F)$ctime))
```

With a README.md created or updated that clearly states the folder/date/time that was used to generate the document.

### Describe alternatives you have considered

none

PR #251, opened 2025-05-21
Make message/.md file describing model_results

sbreitbart-NOAA:
# What is the feature?
* Make message/.md file describing model_results when create_template is used, as per https://github.com/nmfs-ost/asar/issues/201

* Modifying create_template

* No


R/create_template.R
Schiano-NOAA, line 1: This is good temporarily but since we are going to change some of this foundation, it will be deprecated soon

sbreitbart-NOAA: 👍 I'm sure we can carry the idea of this into the revised functions, in some form. It'll be useful to have it as a reminder, at the very least

PR #255, opened 2025-05-21
Finish developing glossary GH action

sbreitbart-NOAA:
# What is the feature?
* GH action that creates a PR on active branch if export_glossary() is altered, signifying that the report_glossary.tex has been changed
* GH action runs on any branch

*

*


k-doering-NOAA:
@sbreitbart-NOAA I  know I said the other day not to worry about the failing call-calc-cov-summary job, but it does look like something may be up...perhaps running r cmd check could illuminate if it is something wrong with the pkg build or with the summary job itself?

Also, I realized we added `@export` to the glossary function, and I'm wondering if we actually should not have done that - I think it makes sense to leave it as is if users should be able to run it, but if it is for developers only, I think maybe removing @export and accessing it only with `asar:::`  (three colons as opposed to 2, `asar::`) may be better. An intermediate option is leaving it exported, but then using `@keywords internal` to export it but minimize the documentation that users would see, as suggested by [the exporting section of r pkgs](https://r-pkgs.org/dependencies-in-practice.html#what-to-export)

sbreitbart-NOAA:
> @sbreitbart-NOAA I know I said the other day not to worry about the failing call-calc-cov-summary job, but it does look like something may be up...perhaps running r cmd check could illuminate if it is something wrong with the pkg build or with the summary job itself?

@k-doering-NOAA I believe the r cmd check is failing because of a test that's failing. We're planning to update the code that will fix this test in the near future, so it should be fixed, and then I think call-calc-cov-summary will be fixed too. I'll keep an eye on this and let you know if it still isn't resolving.

> Also, I realized we added `@export` to the glossary function, and I'm wondering if we actually should not have done that - I think it makes sense to leave it as is if users should be able to run it, but if it is for developers only, I think maybe removing @export and accessing it only with `asar:::` (three colons as opposed to 2, `asar::`) may be better. An intermediate option is leaving it exported, but then using `@keywords internal` to export it but minimize the documentation that users would see, as suggested by [the exporting section of r pkgs](https://r-pkgs.org/dependencies-in-practice.html#what-to-export)

Since we're putting the files necessary to run the glossary function into .Rbuildignore, then the user won't be able to run the function. Good call! I'll remove @export from the function. Thanks!

k-doering-NOAA:
Sounds good!

Also - I wouldn't worry *too* much about .Rbuildignore, since it is one function that generates 1 file that isn't that large) - not a big deal if users get it! But up to you all. The GHA may need some modification if the function itself is put into .Rbuildignore

PR #257, opened 2025-05-21
Address assigned items in "format_quarto" section of cleanup plan

sbreitbart-NOAA:
# What is the feature?
* Comment out option and add stop when docx is selected (not currently functional)
* Set defaults in replace of NULL in all functions

* see above

* No


Schiano-NOAA (approved):
Perfect!

Issue #260, opened 2025-05-23
Organizing Functions and Major Changes


sbreitbart-NOAA:
# Organizing Functions and Major Changes

## Sam will work on create\_template first; will ping Sophie when ready to hand off

\* make branch off of **cleanup-pkg** – has some current changes \*

# General

https://github.com/nmfs-ost/asar/pull/256
https://github.com/nmfs-ost/asar/pull/278
  - [x] @Schiano-NOAA Review unnecessary arguments from functions @ end  
  - [x] Replace duplicated arguments from subfunctions with “...” @ end  
  - [x] @sbreitbart-NOAA *Do after previous task:* Organize function arguments based on priority (check @ end)  
  - [x] Documentation (and check function args too)  
  - [x] Set defaults in replace of NULL in all functions (check @ end) (- No more nulls\! Species is “species”, year \= (current year in the code), etc.)
  - [x] @sbreitbart-NOAA Add disclaimer into outline
  - [x] @sbreitbart-NOAA Remove start-up message (remove zzz.R)
  - [x] @sbreitbart-NOAA Add docs folder to .Rbuildignore **← this was already in there**
  - [x] **Update associated examples and tests for each function changed** (check @ end)  
  - [x]  Substitute cli in interface for messages and such throughout package (leave open for now)  
  - [x] @Schiano-NOAA Change tutorial @ end  
  - [x] Review documentation @ end  
  - [x] Update/add tests @ end  
  - [ ] remove commented out code to clean up and make more straightforward
 
# @Schiano-NOAA `convert_output`  

  - [x] Remove options for save
  - [x] Remove outdir as arg
  - [x] Change output file so reads in file path  
  - [x] Add fleet\_names as option for converter
  - [x]  Extract fleet names from ss3 
  - [x] Expand ss3 “KEYWORDS”  
  - [x] Fix error with BAM identified by Nikolai  
  - [x] Set defaults in replace of NULL in all functions (where applicable)
  - [x] Check the regex in the prev\_results \<- stringr::str\_replace{..} (line 884\)  
  
# `create_template`  

  - [x] @Schiano-NOAA ~~Remove argument “new\_template” \- in place look for a current skeleton in the file dir or alter so there is prev\_report or something like that for it to get copied over- Sam rethinking this rn~~
  - [ ] #293
            - For the latter, will need to add the functioning  
  - [x] @sbreitbart-NOAA Remove complex argument (not used)  
[...]

PR #264, opened 2025-05-28
Cleanup pkg create figures doc

sbreitbart-NOAA:
Left to do:
I think these changes should be made after this PR is merged because these changes will affect 1) many files in asar, including create_template, and 2) many files in stockplotr. I'd prefer to make widespread changes to stockplotr when we're closer to merging these changes into dev, let alone main.
- [ ] change `rda_files` --> `plots` throughout asar
- [ ] change `rda_dir` --> `plot_dir` throughout asar
- [ ] change `rda_files` --> `plots` throughout stockplotr
- [ ] change `rda_dir` --> `plot_dir` throughout stockplotr
- [ ] Update test and add more (at end)

NOTE: I have not updated the function so that it pulls from a "figures" folder. I think we should avoid that, if possible, to keep all figures/tables together in one folder. This function can differentiate between the types if "figure" or "table" is in the filename.

# What is the feature?
* Tasks as per https://github.com/nmfs-ost/asar/issues/260

* Large-scale changes to create_figures_doc()

* Yes

### DIFF
--- a/R/create_figures_doc.R
+++ b/R/create_figures_doc.R
@@ -104,7 +104,7 @@ if (file.exists(file.path(rda_dir, '", fig, "'))){\n
     }
 
     if (length(file_fig_list) == 0){
-      message(paste0("No figure files were found in '", fs::path(rda_dir, "rda_files"), "'."))
+      message(paste0("Note: No figure files were present in '", fs::path(rda_dir, "rda_files"), "'."))
       figures_doc <- "## Figures {#sec-figures}"
     } else {
       # paste rda figure code chunks into one object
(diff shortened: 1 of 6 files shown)


sbreitbart-NOAA:
> This is great! I think this will be a huge improvement and good for long term maintenance. My one suggestion would be to change the working of the messages when there are no non-rda or no rda figures because I thought initially that something was going wrong.
> 
> A potential change could be something like "Note:" in the beginning or saying no figures files in a non-rda format were present" (I think this assumes that they weren't found and can be both a warning and indicator when you know they aren't present)
> 
> Reminder to rerun devtools::document because the documentation is not up-to-date 😄

Done 👍 Thanks Sam!

PR #265, opened 2025-05-28
Automate ss3 keywords

Schiano-NOAA:
# What is the feature?
* automatically pulls keywords that are used in the ss3 report file rather than manually - effort to make the function more robust

* Remove the manual list of keywords and extract them from the Report file similar to `ss3_extract_df()`
* add in new factors to final df
* add in new keywords to groupings and reorganize groupings as needed
* added in keywords for rand

* yes


Schiano-NOAA:
> @Schiano-NOAA could you please resolve the merge conflicts before I review this? I'm not sure which option is the most up-to-date code

I did this earlier, but looks like something got lost or not pushed correctly. Everything should be up to date now!

sbreitbart-NOAA (commented):
Wow, this looks like it was a ton of work! Everything went smoothly when I converted 4 Report.sso files and then plotted the results with stockplotr. The only unexpected thing was the lack of a "finished!" message.

PR #267, opened 2025-06-02
Cleanup pkg create tables doc

sbreitbart-NOAA:
# What is the feature?
* Update create_tables_doc() similarly to create_figures_doc() as per https://github.com/nmfs-ost/asar/issues/260
_Note: only rdas will be supported, not image files. Tables in image format are not machine-readable and pose a barrier to accessibility._

* 

* Yes

### DIFF
--- a/R/create_tables_doc.R
+++ b/R/create_tables_doc.R
@@ -47,10 +47,7 @@ create_tables_doc <- function(subdir = getwd(),
   # set landscape page width (in)
   landscape_pg_width <- 8
 
-  if (include_all == FALSE) {
-    # add option for only adding specified tables
-    warning("Functionality for adding specific tables is still in development. Please set 'include_all' to true and edit the 08_tables.qmd file to remove specific tables from the report.")
-  } else {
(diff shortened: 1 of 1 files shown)


R/create_tables_doc.R
```diff
@@ -20,23 +29,34 @@
...
+  # specify essential_columns for stockplotr-created tables.
 
   # set portrait page width (in)
   portrait_pg_width <- 5
 
   # set landscape page width (in)
   landscape_pg_width <- 8
 
-  if (include_all) {
+  if (include_all == FALSE) {
```
Schiano-NOAA, line 50: An alternative to this is to set this to 
```
if (!include_all) stop("Functionality for adding specific tables is still in development. Please set 'include_all' to true and edit the 08_tables.qmd file to remove specific tables from the report.")
```

then remove the surrounding brackets in the rest of the else statement

Schiano-NOAA (approved):
Looks good! I suggested one minor change that does not impact its function. Feel free to add or ignore

PR #272, opened 2025-06-05
Substitute messages, warnings, and stops with cli

sbreitbart-NOAA:
# What is the feature?
* Substitute messages and warnings with cli package versions

* see above

* No

### DIFF
--- a/R/add_alttext.r
+++ b/R/add_alttext.r
@@ -64,7 +64,7 @@ add_alttext <- function(
     compile = TRUE,
     rename = NULL) {
   # Read latex file
-  if (!file.exists(file.path(dir, x))) cli::cli_abort(glue::glue("File {dir}/{x} does not exist!"))
+  if (!file.exists(file.path(dir, x))) cli::cli_abort("File {dir}/{x} does not exist!")
   tex_file <- readLines(file.path(dir, x))
 
   # Check: count instances of pattern
--- a/R/add_alttext.r
+++ b/R/add_alttext.r
@@ -145,9 +145,7 @@ add_alttext <- function(
         dplyr::pull(alt_text)
       if (is.na(label_line)) {
         alttext_i <- ""
-        warning(glue::glue(
-          "No alternative text found for {line_label}."
-        ))
+        cli::cli_alert_warning("No alternative text found for {line_label}.")
       }
       # Add selected alttext onto end of the line
       tex_file[i] <- gsub(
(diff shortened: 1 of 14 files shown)


sbreitbart-NOAA:
@Schiano-NOAA I didn't change create_template because you're still actively updating that function

R/add_alttext.r
```diff
@@ -77,11 +77,11 @@ add_alttext <- function(
   #   )
 
   # Check if alt text csv is where indicated
-  if (!file.exists(file.path(alttext_csv_dir, "captions_alt_text.csv"))) stop(glue::glue("'captions_alt_text.csv' not found in {alttext_csv_dir}."))
+  if (!file.exists(file.path(alttext_csv_dir, "captions_alt_text.csv"))) cli::cli_abort("'captions_alt_text.csv' not found in {alttext_csv_dir}.")
```
Schiano-NOAA, line 80: I think this needs glue::glue

R/add_alttext.r
```diff
@@ -145,9 +145,7 @@ add_alttext <- function(
         dplyr::pull(alt_text)
       if (is.na(label_line)) {
         alttext_i <- ""
-        warning(glue::glue(
-          "No alternative text found for {line_label}."
-        ))
+        cli::cli_alert_warning("No alternative text found for {line_label}.")
```
Schiano-NOAA, line 148: glue::glue

R/add_alttext.r
Schiano-NOAA, line 80:
```suggestion
  if (!file.exists(file.path(alttext_csv_dir, "captions_alt_text.csv"))) cli::cli_abort(glue::glue("'captions_alt_text.csv' not found in {alttext_csv_dir}."))
```

sbreitbart-NOAA: It should work without glue: https://cli.r-lib.org/reference/inline-markup.html#collapsing-inline-vectors

Schiano-NOAA (approved):

kellijohnson-NOAA:
I know that I am late to the game and that this PR has already been merged into cleanup-pkg but I am wondering how I can be the most helpful with regard to additional formatting that is available from {cli} that is not being taken advantage of. For example, with a file path one should use the inline styler, e.g., `cli::cli_abort("{.file {my_file_path}} is a bad file"}` which creates a clickable link to the file. Also, there is no need to use `glue::glue()` inside of a cli function.

PR #273, opened 2025-06-10
Update bsb_conout.csv

sbreitbart-NOAA:
# What is the feature?
* Replace bsb_conout.csv with new version created with updated convert_output() function

* see above

* No


Schiano-NOAA:
That's what I'm thinking just to not add more info to the package otherwise looks good to me

PR #278, opened 2025-06-11
General Review and final adjustments to documentation for cleanup-pkg

Schiano-NOAA:
# What is the feature?
* update documentation and order arguments by priority for all user facing functions

* move around arguments

* no


Schiano-NOAA:
@sbreitbart-NOAA I requested review if you wanted to do a final check but I will resolve the merge conflicts and merge in if you're good with that. The merge conflicts will pretty much be resolved because the updated documentation on this branch will supersede that on other branches

R/add_authors.R
```diff
@@ -0,0 +1,142 @@
...
+#' \dontrun{
+#' add_authors(
+#'   author = c("Danny Phantom"="SWFSC-LJCA","John Snow"="AFSC-ABL","Jane Doe"="NWFSC-SWA"),
+#'   rerender_skeleton = FALSE
+#' )
+#' }
+add_authors <- function(
+    author,
+    rerender_skeleton = FALSE,
+    prev_skeleton = NULL) {
```
sbreitbart-NOAA, line 19: From R cmd check: `Undocumented arguments in Rd file 'add_authors.Rd'
    'prev_skeleton'`

This arg isn't currently defined in create_template anymore so it's not inherited

sbreitbart-NOAA (commented):
Just had a couple of comments. Otherwise, looks good to merge!

R/add_authors.R
Schiano-NOAA: Ahh it actually was never an argument in create_template to begin with so just an oversight on my part. I'll add documentation. Good catch!

PR #290, opened 2025-06-26
Add params chunk to outlne

Schiano-NOAA:
# What is the feature?
* Translate the parameters loaded in the yaml to a chunk for shorter in-text references for the user

* added new chunk to the outline with parameters loaded as objects

* no

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -644,6 +644,20 @@ create_template <- function(
 
       if (!rerender_skeleton) cli::cli_alert_success("Built YAML header.")
 
+      ##### Params chunk ----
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA (approved):
Looks good. Can you please remind me about the purpose of this chunk?

Schiano-NOAA:
@sbreitbart-NOAA This reduces the amount users have to type. If they want to use the parameterization from quarto in the yaml they need to put this `r params$species` as in-line text. The chunk allows users to now do this `r species` instead which is just less characters and easier to remember imo

PR #291, opened 2025-06-27
Remove 'eval' from figures and tables docs

sbreitbart-NOAA:
# What is the feature?
* Remove 'eval' from figures and tables docs
* Fix other bugs related to creating figures/tables docs

* 

* No


Schiano-NOAA (commented):
Looks good! One minor change to the Figures header to reduce to the same header level as the rest of the files and then good to go

+rebase, I just merge in the params chunk PR

PR #295, opened 2025-06-27
Option to save in convert_output

Schiano-NOAA:
# What is the feature?
* add back in save option to convert_output -- default is now rda

* save when save_dir is not NULL

* no

### DIFF
--- a/R/convert_output.R
+++ b/R/convert_output.R
@@ -2,7 +2,7 @@
 #'
 #' Format stock assessment output files to a standardized format.
 #'
-#' @param output_file Assessment model output file path
+#' @param file Assessment model output file path
 #' @param model Assessment model used in evaluation ("ss3", "bam",
 #'  "asap", "fims", "amak", "ms-java", "wham", "mas").
 #' @param fleet_names Names of fleets in the assessment model as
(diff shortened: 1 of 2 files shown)


R/convert_output.R
```diff
@@ -8,27 +8,32 @@
...
 #'  indicate the fleets names as an acronym in a vector
+#'  @param save_dir File path to save the converted output file.
 #'
 #' @author Samantha Schiano
 #'
 #' @return A reformatted and standardized version of assessment model results
 #'         for application in building a stock assessment reports and to easily
 #'         adapt results among regional assessments. The resulting object is
 #'         simply a transformed and machine readable version of a model output file.
+#'         Converted data frame is always out. It will also be saved if save_dir
```
sbreitbart-NOAA, line 19:
```suggestion
#'         Converted data frame is always `out_new`. It will also be saved if save_dir 
```

R/convert_output.R
```diff
@@ -8,27 +8,32 @@
...
 #'
 #' @examples
 #' \dontrun{
 #' convert_output(
-#' output_file = "~/Documents/ss3_models/model1/Report.sso",
+#' output_file = here::here("model1", "Report.sso"),
 #' model = "ss3",
-#' fleet_names = c("TWL", "NONTWL"))
+#' fleet_names = c("TWL", "NONTWL"),
+#' save_dir = here::here("standard_output.rda"))
```
sbreitbart-NOAA, line 30: I think `save_dir` is confusing because the user is expected to provide a filename, not just a directory for the newly converted file. I think a better argument name would be `converted_file_name`, and I'd have the function add on the ".rda" automatically. Even better, (this would be a breaking change), I'd consider changing `output_file` to something like `input`, and `save_dir` to `output`, because it's confusing when your "output file" is actually the function's input file.

PR #315, opened 2025-07-09
Remove latex warnings on tagging

Schiano-NOAA:
# What is the feature?
* Remove latex warning when tagging document

* removed notation for one package that gets loaded in from quarto
* add in soul package

* no

### DIFF
--- a/R/add_tagging.r
+++ b/R/add_tagging.r
@@ -72,6 +72,13 @@ add_tagging <- function(
       tex_file[issue_color]
     )
   }
+  # Comment out microtype package
+  microtype_line1 <- grep("\\IfFileExists\\{microtype.sty\\}\\{\\% use microtype if available", tex_file)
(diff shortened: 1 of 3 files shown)


sbreitbart-NOAA (commented):
The function works when compile = F. When it's true, I get this error:

! LaTeX Error: No support files for \DocumentMetadata found.

I don't think this error is due to these changes, but just wanted to let you know

Schiano-NOAA:
@sbreitbart-NOAA are you running it when your working directory is where the tex file is? I tried both ways and they worked ok

sbreitbart-NOAA:
That's what I did (`tinytex::install_tinytex(bundle = "TinyTeX-2")`) 🙂 it's taking a while to compile but I'm taking that as a good sign!

sbreitbart-NOAA:
It worked! Will add this to the FAQs

PR #320, opened 2025-07-09
Add TOC to pdf bookmarks

sbreitbart-NOAA:
# What is the feature?
* Add TOC to pdf bookmarks, in response to Brian's question ("Is there a way to tag the Table of Contents so that it appears within the bookmarks section of the document? Right now the first bookmark I have is the Executive Summary.")

* Adding a line to a latex partial

* No


Schiano-NOAA (commented):
I am not sure if this is something everyone will want so let's hold off merging for now

PR #328, opened 2025-07-16
Github action fixes

k-doering-NOAA:
# What is the feature?
* Installs quarto for the r cmd check action (and a few small config changes)
* Removes tests from .Rbuildignore, because the tests were being skipped on r cmd check when included in the .Rbuildignore file

* Edits to github action file
* Edits to .Rbuildignore

The edit to r build ignore should help with the code coverage summaries and badge workflows failing by allowing tests to be run. However, it looks like the test are failing on linux only (I also observed this on codespaces), so the workflows won't succeed just yet.

I wasn't sure if we should continue trying to troubleshoot the tests in this branch or get this merged in and then try to troubleshoot.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -13,4 +13,3 @@
 ^inst/glossary/raw_acronym_lists$
 ^\.octocov\.yml$
 ^\.devcontainer$
-^tests$


Schiano-NOAA:
I think good to merge!

k-doering-NOAA:
@sbreitbart-NOAA , I think the code coverage badge and summary ones are failing because they run on linux - and the [linux r cmd check](https://github.com/nmfs-ost/asar/actions/runs/16348603047/job/46189017696) is the one failing (mac and windows are working, like you saw!) It does look like the tests are running now, though, so it is getting closer!

Let me know if you would like help debugging, as sometimes fixing failing tests on one OS can be difficult - it's sometimes hard to discover the root cause.

sbreitbart-NOAA:
@k-doering-NOAA I'd definitely appreciate your help debugging this. I've tried opening a codespaces to investigate myself and it keeps crashing

PR #335, opened 2025-07-21
Use {goodpractices} package and start resolving things pointed out in that

e-perl-NOAA:
Part of NSAP code cleanup

# What is the feature?
* Turn -> into <- 
* convert sapply to vapply

* Similar stuff may come up when using {goodpractices} for stockplotr


sbreitbart-NOAA:
Note: `find.package("asar")` found the true location of `asar`.

Schiano-NOAA:
Changes needed to branch before merge - see #356

PR #339, opened 2025-07-22
try to update codespaces config

k-doering-NOAA:
# What is the feature?
* Update the codespaces config so that (hopefully) it doesn't crash. In addition, removing some unneeded features could speed up the Codespace spin up time.

Testing needed: If @sbreitbart-NOAA or @Schiano-NOAA could try opening a codespace on this branch and using it for a few ASAR development tasks, that would help confirm that the config works!

* Removed c++ compiling tools, because I don't think they are needed here
* added R packages that ASAR imports (so they don't need to be installed during the session). Some are not listed explicitly because tidyverse ones are already included in the image being used.
* Streamlined the rstudio IDE option (if everyone developing is comfortable in vscode, this could be removed)
* Removed unneeded extensions and options. Some used to be needed, but the images have been improved so less customization is needed to use R in codespaces.

* No

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,9 @@
    	"image": "ghcr.io/rocker-org/devcontainer/tidyverse:4.3",
 
 	"features": {
-		"ghcr.io/rocker-org/devcontainer-features/quarto-cli:1": {},
+		"ghcr.io/rocker-org/devcontainer-features/quarto-cli:1": {
+			"installTinyTex": true
+		},
 		"ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
             	"packages": "qpdf"
 		},


sbreitbart-NOAA:
@k-doering-NOAA Thank you for initiating this PR! The changes you made seem logical (I'm not an expert on this, so take that with a grain of salt!)
I've tried opening codespaces on this branch and it still crashes before I can explore. I've tried various changes and still nothing works. Maybe @Schiano-NOAA will be able to make it work?
image

k-doering-NOAA:
hmm, @sbreitbart-NOAA that's an interesting error I haven't seen before - if you go to https://github.com/codespaces and click "use this template" under the "Blank" codespace template, do you get the same result? The error implies that there is no internet access for your computer, but I'm assuming that it is on the internet?

screenshot of blank codespaces template

k-doering-NOAA:
Thanks, Sam - there isn't a rush here, but it would be helpful to know your result when you have a chance.

Schiano-NOAA:
@k-doering-NOAA I was able to successfully start up a codespace off this branch! I made a small change to a file in the most recent commit and was able to work off of there

sbreitbart-NOAA:
Yeah, I'm curious if my issue might change or resolve once I receive my NOAA laptop... if not, I can try to investigate more. But if this works for Sam, that's a very good sign

Schiano-NOAA:
@k-doering-NOAA Okay so devtools::load_all() worked and I tried out running create_template() which also worked but I could not render because tinytex is not installed. Do you think you could add tinytex bundle = "TinyTeX-2" into the default code space as well?

k-doering-NOAA:
good catch! The quarto feature includes an option to install tinytex, which I changed to true. Could you see if that works?

k-doering-NOAA:
Great! Yes, I think it is ready to merge in, when you would like, Sam :)

sbreitbart-NOAA:
Just following up- was able to open a codespaces on the main branch on my new NOAA laptop, so I'm guessing the issue was with my clunky old personal computer 🥲

Issue #340, opened 2025-07-22
[Bug]: R cmd check failing on ubuntu


k-doering-NOAA:
### Describe the bug

R cmd check is failing on ubuntu, but not on other operating systems.

The root of this appears to be one failing test on ubuntu:

```r
── Failed tests ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error (test-ID_tbl_width_class.R:11:3): Table widths calculated correctly
Error in `x[[property]]$data[i, j] <- value`: number of items to replace is not a multiple of replacement length
Backtrace:
    ▆
 1. └─stockplotr::table_indices(dat, make_rda = TRUE, tables_dir = getwd()) at test-ID_tbl_width_class.R:11:3
 2.   └─flextable::align(...)
 3.     └─flextable:::set_par_struct_values(...)
```

The error appears to be coming from this [stockplotr call to flextable::align](https://github.com/nmfs-ost/stockplotr/blob/aec7c79b98067b48b59f793754cb4c7277e30d3c/R/table_indices.R#L137)

On windows 11 (locally for me), the error does not show up, but there is a warning that appears at the same line of the test script.

```r
Warning (test-ID_tbl_width_class.R:11:3): Table widths calculated correctly
There was 1 warning in `dplyr::mutate()`.
i In argument: `year = as.numeric(year)`.
Caused by warning:
! NAs introduced by coercion
Backtrace:
    ▆
 1. └─stockplotr::table_indices(dat, make_rda = TRUE, tables_dir = getwd()) at test-ID_tbl_width_class.R:11:3
 2.   └─stockplotr::write_captions(dat = dat, dir = tables_dir, year = end_year)
 3.     ├─dplyr::mutate(...)
 4.     └─dplyr:::mutate.data.frame(...)
 5.       └─dplyr:::mutate_cols(.data, dplyr_quosures(...), by)
 6.         └─dplyr:::signal_warnings(warnings_state, error_call)
 ```

### Describe how to reproduce the bug

Using the main branch on a ubuntu machine, run

```r
devtools::test(filter = "ID_tbl_width_class")
```

PR #344, opened 2025-07-31
Use simpler table to resolve failing ubuntu tests

sbreitbart-NOAA:
# What is the feature?
* use sample dataset not derived from actual model output for ID_tbl_width_class tests, as per https://github.com/nmfs-ost/asar/issues/340, to resolve failing ubuntu tests

* Using simpler tables

* No


Schiano-NOAA (approved):
💯

Issue #349, opened 2025-08-13
[Bug]: `add_authors()` should fail fast and informatively if author is an unnamed vector


kellijohnson-NOAA:
### Describe the bug

https://github.com/nmfs-ost/asar/blob/d3070f349f12a49ad6e6ae10aa2865de104b5581/R/add_authors.R#L22

leads to `authors_names` being null if someone passes an unnamed vector to `author`, which is exactly what I did. This is particularly problematic because the package behavior has changed and the error message that I received was not helpful at all. Instead, I had to figure out what function it was coming from and debug that function.
FYI
I was getting
```
Error in data.frame(name = author_names, office = author, row.names = NULL) :    
  arguments imply differing number of rows: 0, 1
```

This could be fixed by ensuring that every element passed to author has a name.

And, as a side note, it is counter intuitive that the argument name is singular, i.e., author, but it accepts a vector with a length greater than one.

### Describe how to reproduce the bug

```
add_authors(author = "Kelli Johnson")
  Error in data.frame(name = author_names, office = author, row.names = NULL) :
  arguments imply differing number of rows: 0, 1
```

### Which OS are you seeing the problem on?

Windows

### Which version of asar are you seeing the problem on?

1.4.0

PR #354, opened 2025-08-14
Fix unclear error message when "author" isn't formatted properly

sbreitbart-NOAA:
# What is the feature?
* Fix bugs mentioned in https://github.com/nmfs-ost/asar/issues/349

* Change `author` to `authors` argument to be clearer that multiple authors can be on one report
* Add clear message about cause and solution for error if user doesn't add both author name and affiliations to `authors`

* No

### DIFF
--- a/R/add_authors.R
+++ b/R/add_authors.R
@@ -18,15 +18,19 @@ add_authors <- function(
     authors,
     rerender_skeleton = FALSE,
     prev_skeleton = NULL) {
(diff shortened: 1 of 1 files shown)


Schiano-NOAA (commented):
This looks good and thanks for making these improvements! Before approving, I am wondering if it's worth instead of having it fail, providing a message that there is no affiliation provided with the author and instead it fills in a blank affiliation template like we do when no author is present at all.

I think we still keep the error message, but give a default so it doesn't just fail.

PR #362, opened 2025-08-22
Vignette about adding alt text/captions if non-stockplotr plots used in report

sbreitbart-NOAA:
# What is the feature?
* Vignette about adding alt text/captions if non-stockplotr plots used in report

* No

### DIFF
--- a/vignettes/caps-alttext-alt-workflows.Rmd
+++ b/vignettes/caps-alttext-alt-workflows.Rmd
@@ -21,9 +21,9 @@ This vignette will be updated as we encounter more varieties of customized workf
 
 As explained in the `stockplotr` ["How captions and alternative text are generated" tutorial](https://nmfs-ost.github.io/stockplotr/articles/how-caps-alttext-are-made.html), exporting at least one figure or table will produce a 'captions_alt_text_template.csv' file, which contains the templates for each table or figure's caption/alternative text. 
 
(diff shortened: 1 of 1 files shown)


vignettes/caps-alttext-alt-workflows.Rmd
```diff
@@ -0,0 +1,34 @@
...
+
+This vignette will be updated as we encounter more varieties of customized workflows.
+
+# When you've created plots without `stockplotr`
+
+As explained in the `stockplotr` ["How captions and alternative text are generated" tutorial](https://nmfs-ost.github.io/stockplotr/articles/how-caps-alttext-are-made.html), exporting at least one figure or table will produce a 'captions_alt_text_template.csv' file, which contains the templates for each table or figure's caption/alternative text. 
+
+* If you've created **all of your tables and figures without `stockplotr`**, then you will not have a version of this csv at all. Move onto the [Create new template file](#sec-create-new-template) section.
+
+* If you've created **at least one table/figure without `stockplotr`**, then *you will have a version of this csv*, but you will be missing the caption/alt text for the plots made without `stockplotr`. Move onto the [Add to existing template file](#sec-add-to-template) section.
```
Schiano-NOAA, line 26: I would suggest changing the wording of "move onto" to "navigate to"

Schiano-NOAA (approved):
This is very helpful! If we want to add any more vignettes in the future, I think we should think hard about where it will go. The vignette's list on the page is getting lengthy so it might be good to brainstorm a better way to present the information rather than a single drop down with a list of vignettes.

Issue #365, opened 2025-08-27
[Feature]: improve how create_figures/tables_doc adds plots


sbreitbart-NOAA:
### Is your feature request related to a problem? Please describe.

Edit `create_figures_doc()`/`create_tables_doc()` so that if the text identifying the tables/figures qmd as empty isn't present, then the new figures/tables code is appended, rather than used to overwrite the existing text. This would ensure that someone's own table/figure code isn't overwritten if they've already added that code first, for instance.

Text identifying tables doc as empty:
> Please refer to the `stockplotr` package downloaded from remotes::install_github('nmfs-ost/stockplotr') to add premade tables.

### Describe the solution you would like.

see above

### Describe alternatives you have considered

see above

PR #366, opened 2025-08-28
Append plot code

sbreitbart-NOAA:
# What is the feature?
* Ability to append figure/table chunks, and issue warnings for repeated labels, described originally in https://github.com/nmfs-ost/asar/issues/365

* Updating create_tables/figures_docs() and tests

* No

### DIFF
--- a/R/create_figures_doc.R
+++ b/R/create_figures_doc.R
@@ -18,7 +18,26 @@
 #' }
 create_figures_doc <- function(subdir = getwd(),
                                figures_dir = getwd()) {
-  figures_doc_header <- "# Figures {#sec-figures}\n \n"
+  
+  # append figure-producing code to non-empty figures doc, if it exists, vs. overwriting it
+  append <- FALSE
+  if (length(file.path(subdir, list.files(subdir, pattern = "figures.qmd"))) == 1) {
+    existing_figs_doc <- file.path(subdir, list.files(subdir, pattern = "figures.qmd"))
+    figure_content <- readLines(existing_figs_doc) |>
+      suppressWarnings()
+    empty_doc_text <- "Please refer to the `stockplotr` package downloaded from remotes::install_github('nmfs-ost/stockplotr') to add premade figures."
+    if ((empty_doc_text %in% figure_content)){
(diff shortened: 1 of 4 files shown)


R/create_figures_doc.R
```diff
@@ -18,7 +18,26 @@
...
+  
+  # append figure-producing code to non-empty figures doc, if it exists, vs. overwriting it
+  append <- FALSE
+  if (length(file.path(subdir, list.files(subdir, pattern = "figures.qmd"))) == 1) {
+    existing_figs_doc <- file.path(subdir, list.files(subdir, pattern = "figures.qmd"))
+    figure_content <- readLines(existing_figs_doc) |>
+      suppressWarnings()
+    empty_doc_text <- "Please refer to the `stockplotr` package downloaded from remotes::install_github('nmfs-ost/stockplotr') to add premade figures."
+    if ((empty_doc_text %in% figure_content)){
+      cli::cli_alert_info("Empty figures doc will be overwritten to include figures in `figures_dir`.")
```
Schiano-NOAA, line 30: I don't think this is necessary because if they are running create_figures_doc, then it's implied they are adding to it. If it's empty, then there no need to worry  about anything being overwritten

Schiano-NOAA: I didn't get this message when I ran the tables doc and filled it in so maybe I am reading this wrong

Schiano-NOAA (approved):
Looks good! I didn't run devtools::check but the tests ran well and I tried it out and it worked as expected. Thanks for doing this update. A huge improvement to the tool :)

R/create_figures_doc.R
sbreitbart-NOAA: This message would've arisen if you ran `create_tables_doc()` without any tables folder (so it'd create a basically empty doc), then ran `create_tables_doc()` again with a tables folder holding a table. I've removed that message, though.

You *will still* get a (different) message if you have a tables doc that doesn't include the text about how you should make tables with stockplotr, yada yada (so it's not "empty"), and then run `create_tables_doc()` again with a tables folder holding a table, since in that case, you'll be appending tables to a doc that should already have at least one table in it

Issue #355, opened 2025-08-15
[Bug]: convert_output issues list


Schiano-NOAA:
## What's going on?

convert_output has a list of minor issues that would make it function better and general changes it needs. In this issue, I am opening a task list for what needs to be done for its next release.

## Task List

- [x] Add current FIMS output naming conventions to converter
- [x] #305 
- [x] Review why there are differences in the Gulf Red Snapper example output when version is correct
- [x] Run red snapper output in ss_output and test its fxnality
- [x] Optimize SELEX_database to run faster
- [x] Set estimates, uncertainty, year columns to numeric
- [x] Find where rec devs uncertainty is going for BAM 
- [x] Find where rec devs uncertainty is going for SS3
- [x] Align naming conventions for landings/catch for all models
- [ ] Look into approach to index columns (i.e. if subseason then incorporate into season)
- [x] Check for M in all model types ([stockplotr issue #67](https://github.com/nmfs-ost/stockplotr/pull/67))
- [x] Fix naming conventions for spr (not the same between SS3 and BAM)
- [x] Add functionality for running converter on online file
- [ ] Add units column
- [ ] Extract units from model output files
- [x] check on rec devs coming from module NA

## Side Tasks
- [ ] Write general function for extracting quantities in preamble

Schiano-NOAA:
Pushing units column to next release

PR #367, opened 2025-09-03
[Bug]: Improve convert_output based on #355

Schiano-NOAA:
# What is the feature?
* Improvements to the converter from #355
* Removed sexes column
* Added FIMS into converter
* Improved standardization of naming conventions
* Also apparently fixed the issue with creating a custom outline in this branch as well
* Adjusted to allow users to convert files located online rather than just locally

* Adjusted naming and other bugs with extracting mainly fleets and some naming separation specifically for BAM

* No

### DIFF
--- a/R/convert_output.R
+++ b/R/convert_output.R
@@ -1642,14 +1642,16 @@ convert_output <- function(
     cli::cli_abort("JABBA output not currently compatible.")
   } else if (model == "fims") {
     if (grepl(".RDS", file)) {
-      out_new <- readRDS(file)
+      fims_output <- readRDS(file)
     } else {
-      out_new <- file
+      fims_output <- file
     }
     # TEMPORARY -- add in era to df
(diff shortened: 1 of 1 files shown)


Schiano-NOAA:
Left to do:
* Test and fix tests
* Convert all output example files and run them through stockplotr functions for cross-testing
* Document
* Resolve merge conflicts

Schiano-NOAA:
> A couple of minor suggestions, but otherwise this looks good to merge! Note that I haven't tested the compatibility with stockplotr branches. Great job on addressing so many changes to the converter.

Great! Sounds good. I have checked the output with all of our stockplotr files (it was the easiest way to identify issues with the converter :)), so we should be good on that side of things. Thanks! I'll make these changes as soon as I can then merge into dev. I plan to write up release notes and push out the new release today or tomorrow

PR #373, opened 2025-09-15
[Hotfix]: Convert_output in v1.5.0

Schiano-NOAA:
# What is the feature?
* Fix issues identified in the convert_output function that were missed specifically dealing with:
* NAs in the label (this should NEVER be the case)
* changing the name of column len_bins to length_bins
* translating 0, 1, 2, 3 in the sex column to respective names
* expand regex options in case_when where they were extracted improperly
* export converted output as a tibble

* Added new standard names aligned with missing labels in csv for SS3
* Expanded case_when in label column creation to accurately extract labels
* When reading in std names sheet, the empty values are read in as NA (previously it was either NA or "" with no predictable way which this would happen (contributed to empty labels issue)
* set returned object as tibble

* yes, should fix incompatibility issues between asar and stockplotr


R/convert_output.R
```diff
@@ -90,7 +90,8 @@ convert_output <- function(
     bin = NA,
     age_a = NA,
     len_bins = NA,
```
sbreitbart-NOAA, line 92: could you also pls update the DESCRIPTION's version to 1.5?

sbreitbart-NOAA (commented):
Besides my mostly minor comments, everything else looks good! 👍

Schiano-NOAA:
@sbreitbart-NOAA I just finished making the edits you mentioned. I am unable to rerun devtools::document because it says my roxygen is out of date, but it also won't let me download the newest version atm. Could you quickly document and commit to this branch before I merge?

Schiano-NOAA:
> > @sbreitbart-NOAA I just finished making the edits you mentioned. I am unable to rerun devtools::document because it says my roxygen is out of date, but it also won't let me download the newest version atm. Could you quickly document and commit to this branch before I merge?
> 
> `devtools::document()` only demanded one small change in the `create_yaml()` documentation. Were you expecting more changes?

Nope! I thought maybe it would update the version somewhere, but all good. Thanks!

PR #386, opened 2025-10-22
Change likelihood col in converted output to numeric

Schiano-NOAA:
# What is the feature?
* Changing the likelihood column in a converted output to numeric

* as numeric to the column at the end, same process as other cols

* no


Schiano-NOAA:
Note: I ran a local check and some tests failed that were unrelated to this PR. Ignore during review. They should be fixed when {stockplotr} is back up working

sbreitbart-NOAA (approved):

PR #405, opened 2025-12-19
[Hotfix]: Make alttext work for cover image and when pandocbounded is not present

Schiano-NOAA:
# What is the feature?
* Adjusts the method from pdftooltip to pandocbounded for alt text in the cover page
* Instead of removing lines that do not have pandocbounded or tooltip, the function just removes the line for the artifact since this is the only known line there that has alt text

* change format to the alt text in the cover page to match that in add_alttext
* Remove the loop that id's which images don't have pandocbounded and instead id the line that has the artifact

* no

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: asar
 Title: Build NOAA Stock Assessment Report
-Version: 1.6.0
+Version: 1.6.1
 Authors@R: c(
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA (dismissed):
Looks good! I'm not sure why only the mac R Cmd check is failing, but that happened to me last week and eventually it disappeared.

Schiano-NOAA:
@sbreitbart-NOAA great! I'm going to push a check for the artifact then merge after checks

PR #413, opened 2026-01-20
remove with_dir() wrapper from a11y fxns' examples

sbreitbart-NOAA:
# What is the feature?
* remove with_dir() wrapper from a11y fxns' examples (not needed anymore)
* update documentation

* updating examples and tests

* No

### DIFF
--- a/R/add_accessibility.R
+++ b/R/add_accessibility.R
@@ -37,18 +37,13 @@
 #'   section_location = "after-introduction"
 #' )
 #'
-#' path <- getwd()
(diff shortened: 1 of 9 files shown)


Schiano-NOAA:
Could you remind me why we are removing this from the function? By using with_dir, the user doesn't have to set their WD to the report file to render. The function is using the dir in the args. Maybe I'm too deep into this and am misunderstanding/forgetting?

sbreitbart-NOAA:
@Schiano-NOAA I realized the PR title and description didn't clearly say the changes were to the examples, not the code itself. My bad! Updated those things. Will merge.

PR #429, opened 2026-02-25
Update functions that relied on 'flextable' and convert to 'gt'

sbreitbart-NOAA:
# What is the feature?
* Update functions that relied on 'flextable' and convert to 'gt'

* Update ID_tbl_width_class() to work with gt tables, vs flextables
* Create ID_tbl_length_class() to ID table length, to decide whether to split them across pages lengthwise
* Update create_tables_doc() to properly format and split tables with varying widths and lengths in tables qmd
* Update/create tests
* Update render_lg_table()
* Update captions for split tables by removing individual column names shown, and instead state number of table vs total number of tables
* Edit add_child() to add line to tables chunk in skeleton to add  "clearpage" latex command, which ensures all tables will be plotted before any other section starts (otherwise, Figures and Appendix titles appeared between tables)
* Other minor fixes like removing stale, duplicate license

* Now it better connects with stockplotr's table exporting functionality

These reports can be tagged. ✅


Schiano-NOAA:
Ugh yeah still getting the same issue with your code too. We can go through the source code on my machine next Tuesday to see what is going on

Issue #416, opened 2026-01-27
[Bug]: 'output_and_quantities' chunk in skeleton missing line to load in model results


sbreitbart-NOAA:
### Describe the bug

When `create_template()` is run with `rerender_skeleton` = TRUE and the model results are added, the 'output_and_quantities' chunk in the skeleton is missing the line to load in model results and the code below that doesn't work.

### Describe how to reproduce the bug

Example from the [NSAW workshop](https://nmfs-ost.github.io/workflows-workshop/Curriculum/nsaw.html#asarcreate_template):

```
create_template(
    format = "pdf",
    type = "SAR",
    office = "NWFSC",
    region = "U.S. West Coast",
    species = "Petrale sole",
    spp_latin = "Eopsetta jordani",
    authors = c("Samantha Schiano" = "OST", "Jane Doe" = "SWFSC-LJCA"),
    file_dir = getwd()
)

petrale <- stockplotr::convert_output(
    file.path("example_output", "Report.sso"),
    save_dir = file.path(getwd(), "report", "petrale_conout.rda"))

asar::create_template(
    rerender_skeleton = TRUE,
    file_dir = file.path(getwd(), "report"),
    authors = c("John Doe"="NEFSC"),
    model_results = file.path(getwd(), "report", "petrale_conout.rda"),
    custom = TRUE,
    new_section = "Changes to the Model",
    section_location = "after-introduction"
)
```

### Screenshots

The top of the 'output_and_quantities' chunk:

```{r} 
#| label: 'output_and_quantities'
#| warning: false 
#| eval: true 
#| include: false
# load converted output from stockplotr::convert_output() 

# Call reference points and quantities below 
output <-file.path(getwd(), "report", "petrale_conout.rda") |>
  dplyr::mutate(estimate = as.numeric(estimate), 
    uncertainty = as.numeric(uncertainty))
source("preamble.R") 
# Available quantities
start_year
...
```

### Which OS are you seeing the problem on?

Windows

### Which version of asar are you seeing the problem on?

2.0.1

Schiano-NOAA:
I tested this a little and realized that this functionality works properly when you select to replace the preamble. Imo, this makes sense because if there are results already presented and then we don't want to overwrite that. Also there is the assumption that results are present if there are customizations to the preamble and we want to replace the preamble when adding new results.

A future development to improve this would be to allow model_results to replace the current output file when rerendering, but I think this operates as expected.

[...]

PR #430, opened 2026-03-02
Update preamble prompt

sbreitbart-NOAA:
# What is the feature?
* Clarify preamble prompt from `create_template()` as per https://github.com/nmfs-ost/asar/issues/416#issuecomment-3873703343

* Clarifying the prompt

* No

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -890,15 +890,13 @@ create_template <- function(
 
     # extract old preamble if don't want to change
     if (rerender_skeleton) {
-      question1 <- readline("As you are rerendering your report skeleton,
-you may need a new preamble.
-Do you want to keep the current preamble? (Y/N)")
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA:
> Can you change the message so it matches the one in the issue? I think the current language is not accurate. We don't want to say they "may" need a new preamble, but more so it will be overwritten so do you want to keep it or not. I think the language in the issue "Update the preamble to match entered arguments? (Y/N)" is closer to what we want here

The reason why I didn't take the language from the issue is that it will change the direction of the question. Currently, "yes" would entail keeping the preamble, and with the "update the preamble..." question, "yes" would entail changing the preamble. I thought the user may find this confusing if they've already grown accustomed to it, but it's not anything as major as a breaking change- just something to be aware of.

If you're alright with that, I can change the question to the one from the issue 👍

PR #433, opened 2026-03-17
add gt_split into utils

Schiano-NOAA:
# What is the feature?
* adding the `gt_split` function from the `gt` package into asar that includes the fix for the function

* see [gt PR](https://github.com/rstudio/gt/pull/2132)

* no

### DIFF
--- a/R/create_tables_doc.R
+++ b/R/create_tables_doc.R
@@ -266,7 +266,7 @@ load(file.path(tables_dir, '", stringr::str_remove(tab, "_split"), "'))\n
                 "  gt::cols_width(\n",
                 "    everything() ~ pct(20)\n",
                 "  ) |> \n",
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -158,3 +158,253 @@ SS3_extract_fleet <- function(dat, vers) {
...
+#
+#  This file is part of the 'rstudio/gt' project.
+#
+#  Copyright (c) 2018-2026 gt authors
+#
+#  For full copyright and license information, please look at
+#  https://gt.rstudio.com/LICENSE.html
+#
+#------------------------------------------------------------------------------#
+
```
sbreitbart-NOAA: I think we should cut this and just make it clear in the note (which I see you added) about where this code is from, and that the authors deserve credit, we're adapting the code until a fix is merged, etc.

PR #448, opened 2026-04-20
Shorten glossary labels by subbing FSC acronyms in for full names

sbreitbart-NOAA:
# What is the feature?
* Shorten glossary labels by subbing FSC acronyms in for full names, as suggested in NEFSC workshop

* No

### DIFF
--- a/inst/glossary/report_glossary.tex
+++ b/inst/glossary/report_glossary.tex
@@ -1,3 +1,13 @@
+% NOTE ON SURVEYS:
+% To increase findability, surveys may have multiple entries to link informal with formal names.
+% Specifically, there may be more than one 'label' (first curly braces entry) or 'acronym' (second curly braces entry) with the same 'long form' (third curly braces entry).
+% If the first 'label' you find for a given survey seems too long, look for an alternative entry with the same 'long form' but a different 'label' and/or 'acronym'.
+% For example, there are three entries for the 'long form' 'Northeast Ecosystem Monitoring (EcoMon)\_Fall':
+
+%     \newacronym{ecomon fall}{ECOMON Fall}{Northeast Ecosystem Monitoring (EcoMon)\_Fall}
+%     \newacronym{fall nefsc ecosystem monitoring survey}{Fall Northeast Fisheries Science Center Ecosystem Monitoring Survey}{Northeast Ecosystem Monitoring (EcoMon)\_Fall}
+%     \newacronym{nemef}{NEMEF}{Northeast Ecosystem Monitoring (EcoMon)\_Fall}
+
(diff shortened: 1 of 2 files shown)


inst/glossary/report_glossary.tex
```diff
@@ -303,7 +303,7 @@
 \newacronym{fall juvenile fish survey in the southeastern bering sea and gulf of alaska}{Fall Juvenile Fish Survey in the Southeastern Bering Sea and Gulf of Alaska}{EMA Eastern Bering Sea Juvenile Fish\_Fall}
 \newacronym{fall juvenile survey}{Fall juvenile survey}{EMA Eastern Bering Sea Juvenile Fish\_Fall}
 \newacronym{fall maine - new hampshire inshore trawl survey}{Fall Maine - New Hampshire Inshore Trawl Survey}{NH/ME Inshore Trawl Survey\_Fall}
-\newacronym{fall northeast fisheries science center ecosystem monitoring survey}{Fall Northeast Fisheries Science Center Ecosystem Monitoring Survey}{Northeast Ecosystem Monitoring (EcoMon)\_Fall}
+\newacronym{fall nefsc ecosystem monitoring survey}{Fall Northeast Fisheries Science Center Ecosystem Monitoring Survey}{Northeast Ecosystem Monitoring (EcoMon)\_Fall}
```
Schiano-NOAA, line 316: This is still a really long acronym...maybe worth asking/tagging Dan to see if this is reasonable or it needs to be adjusted.

Schiano-NOAA (approved):
I think this looks good. I had one minor comment, but other changes can be made in a different PR.

inst/glossary/report_glossary.tex
sbreitbart-NOAA: It is still long. However, there's a shorter option [here](https://github.com/nmfs-ost/asar/blob/29f0adf4d6aa03fa59e288fe453d608ff3b4d55a/inst/glossary/report_glossary.tex#L623), on line 623. This is an example of how, for the surveys, if the acronyms were >3 words long (e.g., Fall Northeast Fisheries Science Center Ecosystem Monitoring Survey), I added an option (another row) with a duplicate meaning and an acronym that was created by taking the first letter from each word in the meaning (e.g., NEMEF). My rationale was that, if the user found their survey's formal name, if was long, they could find another option with a shorter acronym. I'll admit that this system hasn't been communicated and that there are probably ways to improve this. I can update the documentation to explain this

Schiano-NOAA: Ah okay I see. Maybe it's too much work but could be good to group them together or alternatively create some type notes as comments in the script to direct the user that there is another option elsewhere. I think we might have talked about some form of this before

PR #455, opened 2026-05-01
[Hotfix]: new section works when custom is FALSE

### LINKED ISSUE #452: [Bug]: option "in" for new section location not working
### Describe the bug

For some reason, both the file and the chunk to load in the file is not working when setting section_location = "in-data" for a new section

I thought at the least the section would be made and the chunk would be missing

`asar::create_template(new_section = c("My New Section"), section_location = c("in-data"))`

### Screenshots

_No response_

_No response_

2.0.4

### Additional Context

_No response_

Schiano-NOAA:
# What is the feature?
* Helps resolve issue #452

* add condition in` create_template` that overwrites custom to FALSE when the user sets `new_section` or `custom_sections` and forgets to set custom = TRUE

* no


sbreitbart-NOAA:
Woops, put my comment on the issue instead of the PR 😅 https://github.com/nmfs-ost/asar/issues/452#issuecomment-4361294557

sbreitbart-NOAA (commented):
Looks good to me! I mentioned a different error on the issue itself but understand if you'd prefer to make that a separate PR.

sbreitbart-NOAA (changes requested):
Thanks for working on this issue! Nearly everything looks good. However, when testing the example in the issue, it doesn't render because both of the new sections' code chunks have the same label: example_chunk_1. I think this will confuse users. One way to avoid duplicating labels would be to make the chunk label identical to the section name + 1, or to remove the code chunks from the new sections entirely.

Schiano-NOAA:
@sbreitbart-NOAA oh good call! I'll remove them entirely because they shouldn't be needed.

Schiano-NOAA:
@sbreitbart-NOAA the issue should be fixed now. I'll merge this in after dev and push this to v1.0 branch as a hotfix as well.

PR #463, opened 2026-05-14
[fix]: Remove duplicated acronyms from glossary

sbreitbart-NOAA:
# What is the feature?
* Remove two duplicates from glossary: bmsy and mt

* Updating `export_glossary()` and glossary tex file

* No

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: asar
 Title: Build NOAA Stock Assessment Report
-Version: 2.0.4
+Version: 2.0.5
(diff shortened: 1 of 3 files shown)


Schiano-NOAA (approved):
Looks good! I think this was only introduced during the most recent version, but could you double check it's not in v2.0 branch as well?

sbreitbart-NOAA:
> Looks good! I think this was only introduced during the most recent version, but could you double check it's not in v2.0 branch as well?

Checked it's not in v2.0 👍 Thanks!

PR #466, opened 2026-05-14
[fix] Update create_template() to find start of preamble

### LINKED ISSUE #462: [fix] conflict between preamble label and preamble.R when rerendering template
### Describe the bug

start_line variable in create_template() sees two occurrences of "preamble", one in #| label: 'preamble' and one in source("preamble.R")

```
create_template()
create_template(
  rerender_skeleton = TRUE,
  file_dir = file.path(getwd(), "report"),
  parameters = TRUE,
  param_names = c("fleet_a", "fleet_b"),
  param_values = c("a long fleet name", "b long fleet name")
)
```

### Screenshots

Warning:

```
Warning message:
In start_line:end_line :
  numerical expression has 2 elements: only the first used
```

_No response_

2.0.4 (not before)

### Additional Context

inserted browser() into create_template() and got the following 
Browse[1]> start_line
[1] 65 75

sbreitbart-NOAA:
# What is the feature?
* Fix https://github.com/nmfs-ost/asar/issues/462

* In `create_template()`, update code that looks for the start of the preamble chunk so it finds it correctly and doesn't identify two lines with "preamble"

* No

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: asar
 Title: Build NOAA Stock Assessment Report
-Version: 2.0.6
+Version: 2.0.7
 Authors@R: c(
     person("Samantha", "Schiano", , "samantha.schiano@noaa.gov", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0003-3744-6428")),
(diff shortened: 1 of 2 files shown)


Schiano-NOAA (approved):
It works 💯

PR #470, opened 2026-05-15
[test] Add tests for rending draft and accessible PDFs

sbreitbart-NOAA:
# What is the feature?
* Add tests for rending draft and accessible PDFs
_Note:_ These tests will only run locally, since they rely on the `tinytex` installation. They won't run on GitHub.

* Adding tests for `add_accessibility()` and `create_template()`

* No

### DIFF
--- a/tests/testthat/test-add_accessibility.R
+++ b/tests/testthat/test-add_accessibility.R
@@ -1,65 +1,42 @@
-# test_that("add_alttext() creates .tex file", {
-#   path <- getwd()
-#   # run convert_output for example model
-#   simple_model <- file.path(test_path("fixtures", "ss3_models"), "models", "Simple")
-#   stockplotr::convert_output(
-#     output_file = "Report.sso",
-#     outdir = simple_model,
-#     model = "SS3",
-#     file_save = TRUE,
-#     save_name = "spp_conout"
-#   ) |> suppressMessages() |> suppressWarnings()
-#
-#   # run stockplotr::exp_all_figs_tables()
-#   stockplotr::exp_all_figs_tables(
-#     dat = utils::read.csv(file.path(path, "spp_conout.csv")),
-#     end_year = 2023,
-#     ref_line = "msy",
-#     ref_line_sb = "msy",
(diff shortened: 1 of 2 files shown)


tests/testthat/test-create_template.R
Schiano-NOAA, line 1: Thanks for making this test! I was thinking about adding the same exact thing :)

tests/testthat/test-add_accessibility.R
Schiano-NOAA, line 1: Was this passing for you? Mine failed

tests/testthat/test-add_accessibility.R
```diff
@@ -1,65 +1,44 @@
...
-# })
-#
+test_that("add_accessibility() runs without error", {
+  # don't run on GitHub because tinytex isn't installed in the GitHub testing environment, so rendering will fail
+  skip_on_ci()
+  
+  stockplotr::plot_biomass(dat = stockplotr::example_data,
+                           interactive = FALSE,
+                           make_rda = TRUE)
+  create_template(
```
Schiano-NOAA, line 8: Might be worth the effort to also save the stockplotr::example_data then load it in through model_results when you create it. 

It should be something like `save(stockplotr::example_data, file = "report/my_data.rda")` then `model_results = "my_data.rda"` in `create_template()`

Then append some reference of a preamble quantity in a child doc.

This could be a different test if we want. It will make sure that the reference in the preamble is working as intended

Note: unlink the saved data at the end.

Schiano-NOAA (changes requested):
See my comments in the docs. Thank you for doing this! It was in the back of my mind. The add_accessibility test did not pass for me

tests/testthat/test-add_accessibility.R
sbreitbart-NOAA: Yes, this passes for me. What error messages are you getting?

Schiano-NOAA: Oops! Not sure what happened but they worked now 😄

Schiano-NOAA (approved):

PR #474, opened 2026-05-19
[fix] In skeleton, fix R params chunk's params$species shortcuts

sbreitbart-NOAA:
# What is the feature?
* In skeleton, fix R params chunk's params$species shortcuts, as recommended during NWFSC workshop

* Modifying `create_template()`

* No


Schiano-NOAA (approved):
LGTM!

PR #476, opened 2026-05-22
[feat] Add chunk options to YAML

### LINKED ISSUE #461: Update global R chunk options in YAML
Create global R chunk options in the yaml (create_yaml()), including tbl-pos: ‘h!’, echo = F. Note: might need to make adjustment to add_chunk defaults
-make it a feature (x.y, not x.y.z)

sbreitbart-NOAA:
# What is the feature?
* Add chunk options to skeleton YAML so that 1) tables are always at the top of the page, but not anchored in a way that interferes with the Tables header and 2) code doesn't show

* Add "echo" and "tbl-pos" chunk options into YAML and remove from add_chunk(); update tests

* No

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: asar
 Title: Build NOAA Stock Assessment Report
-Version: 2.1.1
+Version: 2.2.0
 Authors@R: c(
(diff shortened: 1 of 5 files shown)


Schiano-NOAA (approved):
LGTM!

PR #477, opened 2026-05-27
Update glossary to add duplicated but differentiated terms

sbreitbart-NOAA:
# What is the feature?
* Update glossary to add duplicated but differentiated terms
* update a11y vignette to explain changes for duplicated acronyms and survey names

* Terms can share acronyms (e.g., "GOA") as long as the label is different (e.g., "goa-ak" and "goa-se")

* No


DESCRIPTION
Schiano-NOAA, line 1: We can chat more about this next week, but don't increment the version anymore. Each release will have it's own version, but when features are merged in, the current main branch becomes the version-alpha indicating that there  are features available that don't have a release yet. The versioning doc is updated to reflect this! Thanks 😄

sbreitbart-NOAA: 👍 Sounds good!

sbreitbart-NOAA: So should I make this version 2.2.0-alpha?

Schiano-NOAA: Make it 2.2.0.9000 for now. Kelli and I decided to use alpha instead, but using -alpha made the gha fail so i'll have to look into it. They mean the same thing though

PR #485, opened 2026-05-28
[fix] Update 'office' param in YAML to include glossary syntax

### LINKED ISSUE #450: [Feature]: change default office param to include gls structure

I noticed during the workshop that when using the region parameter, it results in the acronym of the region. We need this to be a part of the defined acronyms so formatting is needed for this parameter in particular to add `\gls{region}`

See above

Leaving as is

### Additional context

_No response_

sbreitbart-NOAA:
# What is the feature?
* Update 'office' param in YAML to include glossary syntax, as it's an acronym, as per https://github.com/nmfs-ost/asar/issues/450
* update SWFSC entry in glossary

* Editing `create_yaml()` and glossary

* No

### DIFF
--- a/R/create_yaml.R
+++ b/R/create_yaml.R
@@ -143,7 +143,7 @@ create_yaml <- function(
         yaml <- stringr::str_replace(yaml, yaml[grep("species: 'species'", yaml)], paste("  ", " ", "species: ", "'", species, "'", sep = ""))
       }
       if (length(office) == 1 & !is.null(office) & any(grepl("office: ''", yaml))) {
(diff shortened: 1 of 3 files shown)


Schiano-NOAA (commented):
Have you run a test that when referencing the office param from both the params in the yaml and the R chunk this works?

If so, then LGTM otherwise please test and lmk!

sbreitbart-NOAA:
> Have you run a test that when referencing the office param from both the params in the yaml and the R chunk this works?
> 
> If so, then LGTM otherwise please test and lmk!

Yes, both work 🥳 Thanks!

PR #486, opened 2026-05-28
[feat] Check and fix duplicated figure/table chunk labels

### LINKED ISSUE #451: [Feature]: add check in create_figures/tables_doc to ensure chunk labels are unique

Accidentally re-run create_figures/tables_doc or have to run fxn again and the rda figures/tables that I already have are repeated and I have to do manual work. It would be great if this is automated to check for duplicates and skip or remove these duplicated pieces

option 1:
- add in rda loading chunks
- find labels for all chunks
- id duplicated locations of chunks and remove duplicates
- output warning in console that some chunks were removed

option 2: 
- check if the .rda path is already in a line in the doc
- skip this in the process if it is already found

keeping as is

### Additional context

_No response_

sbreitbart-NOAA:
# What is the feature?
* Create function that checks for duplicated chunk labels in fig/table docs, and if present, comments out duplicates (as per https://github.com/nmfs-ost/asar/issues/451)
* Update tests

* Creating `fix_duplicate_chunks()` function in utils

* No


Schiano-NOAA:
I'm going to tag team on this and make checks for duplicate rda's/images as well and have both issues solved in one. I tested this out and I think this is a good start, but it does a couple of thing

(1) I think the commenting out is a good solution, but I am going to add a check to see if the same image is loaded in. In the case it is, the the chunk will be removed/won't be added. In the case it's not, I think it would be better to rename the label with a 1 or other alternative so it's still accessible to the user

(2) I will adjust the lapply or a loop to go through the chunks and make sure ones that have already been added are not added

Schiano-NOAA:
@sbreitbart-NOAA this PR is now ready for merge!

tests/testthat/out_new.rda
sbreitbart-NOAA, line 1: This should be deleted, right?

Schiano-NOAA: I think so

R/create_tables_doc.R
```diff
@@ -474,22 +505,26 @@ load(file.path(tables_dir, '", stringr::str_remove(tab, "_split"), "'))\n
...
+      cli::cli_ol(c(
+        "a 'tables' folder in {fs::path(tables_dir)}",
+        ".rda files in the 'tables' folder"
+      ))
+      tables_doc <- paste0(
+        tables_doc_header,
+        empty_doc_text
+      )
+    } else {
+      cli::cli_alert("No new tables detected.")
```
sbreitbart-NOAA, line 524: This line is great. Can you please add a similar one for figures?

PR #507, opened 2026-07-02
[feat]: Add a different species image on `rerender_skeleton = TRUE`

Schiano-NOAA:
# What is the feature?
* Users can change the cover image using spp_image when rerender_skeleton = TRUE

* adjusted the path and if statement conditions

* no


sbreitbart-NOAA (approved):
LGTM!

Issue #441, opened 2026-04-07
[Review]: create_template documentation II


Schiano-NOAA:
# Directions

Please comment on the follow topics regarding documentation of the function [create_template](https://github.com/nmfs-ost/asar/blob/main/R/create_template.R):

- [ ] Arguments (model_results through "..." )
- [ ] Value
- [ ] Examples

Think about the perspective of both new and advanced users when reviewing the documentation.

You can access the documentation quickly and easily by running ?asar::create_template in your console. Remember to only focus on the documentation. If you have any feedback regarding the process of create_template, please leave a separate [issue](https://github.com/nmfs-ost/asar/issues/new/choose).

BenWilliams-NOAA:
Suggested edits below:

```
#' @param model_results Path to an `.rda` file containing model output  
#'   processed via `stockplotr::convert_output()', default: 'NULL'.  
#' @param tables_dir Path to directory containing table files. Defaults to working directory.  
#' @param figures_dir Path to directory containing figure files. Defaults to working directory.  
#' @param spp_image Path to a custom species image. If empty, searches `asar` 
#'   resources for a matching species name.  
#' @param bib_file Path to bibliography file (`.bib`). Defaults to `asar_references.bib` in the working directory.
#' @param new_template Logical. If `TRUE` (default), creates a fresh skeleton. 
#'   If `FALSE`, attempts to pull the last saved version.
#' @param rerender_skeleton Logical. If `TRUE`, updates the `skeleton.qmd` 
#'   YAML and structure while preserving existing `.qmd` section files. Default `FALSE`
#' @param custom Default `FALSE`,  if `TRUE`, allows for non-standard sectioning rather than the default for stock assessments in
#'  the region.
#' @param custom_sections List of existing sections to include in
#' the custom template. Note: this only includes sections within
#'  list.files(system.file("templates", "skeleton",
#'  package = "asar")). The name of the section, rather than the
#'  name of the file, can be used (e.g., 'abstract' rather than
#'  '00_abstract.qmd'). If adding a new section, also use
#'   parameters 'new_section' and 'section_location'.
#' @param new_section Names of section(s) (e.g., introduction, results) or
#' subsection(s) (e.g., a section within the introduction) that will be
#' added to the document. Please make a short list if >1 section/subsection
#' will be added. The template will be created as a quarto document, added
#' into the skeleton, and saved for reference. Default NULL.
#' @param section_location Where new section(s)/subsection(s) will be added to
[...]

PR #509, opened 2026-07-07
Combine param_names and param_values into one argument

sbreitbart-NOAA:
# What is the feature?
* Combine param_names and param_values (in create_template() and create_yaml()) as one arg as per @BenWilliams-NOAA 's suggestion in https://github.com/nmfs-ost/asar/issues/441
* update documentation

* No


Schiano-NOAA (approved):
LGTM! I noticed that the office param has \gls command incorporated into it even when the office = NULL. I don't think this would cause an issue but wanted to mention it because I don't know when that was added.

Thanks for doing this!

PR #511, opened 2026-07-10
[feat] Add new template "type" for SEDAR reports

### LINKED ISSUE #457: [Feature]: Set authorship for SEFSC reports to office rather than author

Change convention of authorship when office = "SEFSC" when creating template to reflect current approach until changes to the assessment process are made

When running` create_template()`, office argument dictates authorship for the SEFSC. The authors argument will identify which address gets placed but the author name is replaced by SEFSC.

- [ ] make adjustment to `add_author()` to allow for <1 word in author name so it doesn't break

Have authors set name to "SEFSC" in authors argument in `create_template`

### Additional context

_No response_

### LINKED ISSUE #502: Make "type" for SEDAR reports in `create_template`

sbreitbart-NOAA:
# What is the feature?
* Add new template "type" for SEDAR reports, based off three reports: [SEDAR 72](https://sedarweb.org/documents/sedar-72-gulf-of-mexico-gag-grouper-final-stock-assessment-report_october2021/), [SEDAR 85](https://sedarweb.org/documents/sedar-85-gulf-of-mexico-yellowedge-grouper-final-stock-assessment-report/), [SEDAR 81](https://www.fisheries.noaa.gov/inport/item/71013)
* Update citation-making process so that if type = SEDAR or office = SEFSC, citation is SEDAR-specific

* Updating `create_template()` and `create_yaml()`
* Adding new template for SEDAR to templates folder

* No

### DIFF
--- a/R/create_template.R
+++ b/R/create_template.R
@@ -16,7 +16,9 @@
 #'
 #' Default: "sar" (a NOAA standard "Stock Assessment Report")
 #'
-#' Options: "sar" (Stock Assessment Report), "nemt" (Northeast Management Track), "pfmc" (Pacific Fishery Management Council), "safe" (Stock Assessment and Fishery Evaluation)
+#' Options: "sar" (Stock Assessment Report), "nemt" (Northeast Management Track),
+#' "pfmc" (Pacific Fishery Management Council), "safe" (Stock Assessment and 
+#' Fishery Evaluation), "sedar" (SouthEast Data, Assessment, and Review)
 #'
 #' @param office Regional Fisheries Science Center producing the report.
 #'
@@ -265,21 +267,26 @@ create_template <- function(
       "Pacific Fishery Management Council" = "pfmc",
       "Stock Assessment and Fishery Evaluation" = "safe",
       "Stock Assessment Report" = "skeleton",
+      "SouthEast Data, Assessment, and Review" = "sedar",
+      "SouthEast Data Assessment and Review" = "sedar",
       "sar" = "skeleton",
       "pfmc" = "pfmc",
       "nemt" = "nemt",
       "safe" = "safe",
+      "sedar" = "sedar",
+      "SEDAR" = "sedar",
       {
         type_fxn <- function() {
           selection <- utils::menu(
             title = "Unrecognized template type. Please select an option below: ",
-            choices = c("Default", "Pacific Fisheries Management Council", "Northeast Management Track", "SAFE")
+            choices = c("Default", "Pacific Fisheries Management Council", "Northeast Management Track", "SAFE", "SEDAR")
           )
           type <- switch(as.character(selection),
             "1" = "skeleton",
             "2" = "pfmc",
             "3" = "nemt",
             "4" = "safe",
+            "5" = "sedar",
             {
               "skeleton"
             }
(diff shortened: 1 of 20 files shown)


R/create_template.R
```diff
@@ -462,7 +469,15 @@ create_template <- function(
 
     asar_folder <- system.file("templates", package = "asar")
     # copy files from specific type folder
-    current_folder <- ifelse(rerender_skeleton, subdir, file.path(asar_folder, type))
+    if (rerender_skeleton){
+      current_folder <- subdir
+    } else {
+      current_folder <- file.path(asar_folder,
```
Schiano-NOAA: It's the same? I'm wondering if this is the case, then we remove functionality for type = "SEDAR" and socialize that the standard guidelines match SEDAR reports already and set the authorship part in the yaml to activate when office = "SEFSC" instead.

What assessments did you look at to determine this is the same? With the hesitation we got, I assumed there would be some differences.

Schiano-NOAA (changes requested):
This is looking good, but it makes me wonder if this is needed. If there is no actual template, it might be worth more socializing that they are the same. The PR would then only make changes when office = "SEFSC" for authorship in the yml

R/create_template.R
sbreitbart-NOAA: This suggestion came from the SEFSC workshop, in the context of the authorship difference. I don't think the intent was to provide a report that matched the SEDAR structure since it seems highly different from the SAR.

I can change this PR to remove "type" and instead update authorship when office = SEFSC 👍

Schiano-NOAA: Ah okay that makes this much more clear. This LGTM then. I did not run a test on it so let me do that then I will approve

Schiano-NOAA:
@sbreitbart-NOAA please also make changes to the citation. Idk how involved that will be, but it still inputs author and I think this doesn't align with SEDAR? Not sure though

R/create_template.R
Schiano-NOAA: After thinking on this over night, I think we should do two things
(1) provide the structure for a SEDAR report using type = "SEDAR" and
(2) change authorship so when office = "SEFSC" then the authorshop and the citation only contain SEDAR and not the analyst names

We discussed post workshop about having a SEDAR template and this potentially helping with the transition

sbreitbart-NOAA:
@lisaailloud-NOAA We are adding a report template that emulates a typical SEDAR report structure. Could you please take a look at the qmd files within the 'inst/templates/sedar' folder and let us know how we could improve the report structure, if at all?

lisaailloud-NOAA:
Thanks! The structure matches well with what we've been doing in the Gulf but I looked at the Atlantic side's reports (e.g. [SEDAR92](https://sedarweb.org/documents/sedar-92-atlantic-blueline-tilefish-final-stock-assessment-report/)) and they are quite different. However, I don't think there's a way to make it work for every case.. Has your approach with other regions to include more or include less? I could set up a meeting to discuss this with representatives from all three branches? 

The only thing I would remove in the meantime is "constant catch" in the 06_projections qmd since we can put that under the OFL / ABC header.

sbreitbart-NOAA:
@lisaailloud-NOAA - thank you for your feedback! I can remove the "constant catch" part.

@Schiano-NOAA Since you set up the other templates, do you have any thoughts re: Lisa's comment about how to make a template that works across a region, as this template is currently better suited for the Gulf vs. Atlantic?

Schiano-NOAA:
@sbreitbart-NOAA @lisaailloud-NOAA I think setting up a meeting with reps from all 3 regions would be best. Then we can at least come to an agreement on an outline for SEDAR if the standard guidelines cannot be adopted at this time.

lisaailloud-NOAA:
That sounds good. I'll look through their reports and compare with the guidelines. I'm thinking it might be easier for everyone to move to the standard guidelines rather than trying to make a SEDAR outline that works for all.. but I'll think on it and set up a meeting.

Schiano-NOAA:
@lisaailloud-NOAA Agreed! Maybe this push is needed then for the standard guidelines

sbreitbart-NOAA:
@Schiano-NOAA I put the participants table in the Introduction because there's already a section for it ("List of Participants"). Also made it a chunk with a very simple gt table instead of a markdown table because the former passes the "regularity" a11y check when tagged (related to the table caption being processed accurately).

sbreitbart-NOAA:
Note: this PR is on hold. We are going to organize further meetings to discuss the best way forward

PR #521, opened 2026-08-03
[chore] update versioning GH action

### LINKED ISSUE #468: [chore] update versioning action
This action should be updated to accommodate the new versioning system. I proposed we also only update the cheatsheet version on the monthly releases rather than each version change.

https://github.com/nmfs-ost/asar/blob/2b98d3d47db362c14acc1a45284fc815bae34689/.github/workflows/new-version-checklist.yml#L24-L27

sbreitbart-NOAA:
# What is the feature?
* Update "new version checklist" GH action to run monthly, as per https://github.com/nmfs-ost/asar/issues/468

* See above

* No

### DIFF
--- a/.github/workflows/new-version-checklist.yml
+++ b/.github/workflows/new-version-checklist.yml
@@ -1,31 +1,32 @@
 # Automatically adds a comment containing a reviewer checklist
-# when a new pull request *from dev to main* is opened.
 name: new-version-checklist
 on:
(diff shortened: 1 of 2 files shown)


Schiano-NOAA (commented):
Not sure if you are able to test it but looks good to me. Can you change the version to .9000 before merge? This branch will probably need a rebase

PR #525, opened 2026-08-10
Fix single-component author names

### LINKED ISSUE #493: [Bug]: single author name or initials throw error
### Describe the bug

When a user inputs only a single name (e.g. Joe) or their initials, it throws an error. The `add_authors()` function is looking for 2 or more strings separated by space which I believe is causing the error. If a check is added to make sure first and last are present OR a blank space or just no addition to the last name line was added, this would fix the issue as well.

```
library(asar)
create_template(authors = c("SB" = "SWFSC"))
```

### Screenshots

_No response_

_No response_

_No response_

### Additional Context

_No response_

msupernaw:
Fixes #493 by supporting author names that contain only one component, such as initials ("SB") or mononyms ("Joe").
Previously, citation generation assumed every author name contained separate given and family names, causing create_template() to fail when parsing a single-component name. This change introduces a documented internal helper that:
Preserves single-component names as the family/display name.
Continues formatting conventional multi-part names as before.
Centralizes duplicated citation-name formatting used by create_citation() and template rerendering.
Removes trailing punctuation added by utils::toBibtex() for single-component names.
Regression tests cover initials, mononyms, and mixed author lists. The original reproducer now completes successfully:


Schiano-NOAA (approved):
LGTM! I made a small adjustment. This has been on our plates for a while and I am so glad we were able to get it off from this code cleanup. Thank you for your contribution! I will be merging shortly.

PR #527, opened 2026-08-10
fix: remove extra `.` 

### LINKED ISSUE #494: [Bug]: citation contains double period after title
### Describe the bug

There is a double period after the title and before the affiliation.
```
Schiano, S., and Doe, J. 2026. Status of Crimson jobfish stock in U.S. waters off the coast of Hawaii in 2026. . National Marine Fisheries Service, [CITY], [STATE].  \pageref*{LastPage}{} pp.
```

I have looked into this somewhat and do not know why this is happening. There are other examples where this does not occur. I believe this is happening in the `create_citation` function.

```
create_template(
  species = "Crimson jobfish",
  office =  "PIFSC",
  authors = c("Sam Schiano"= "OST")
)
```

### Screenshots

_No response_

_No response_

_No response_

### Additional Context

When determining the cause of this issue, the following are related functions:
- `create_citation` is responsible for creating the citation and putting it together (I suspected this is where the bug is occurring, but can not determine it when run line by line)
- all inputs to create_citations are either shared with `create_template()` (the function `create_title()` is used in) or made within the create_template() function.
https://github.com/nmfs-ost/asar/blob/68bbe7bddd5dd9bb53443a94141ffefbe2229e3c/R/create_template.R#L1042-L1121
- Running the example above should produce the issue.
- I thought the reason was because a period is added after the title in `create_citation` and maybe one is made in the string in [`create_title`](https://github.com/nmfs-ost/asar/blob/main/R/create_title.R), but this is NOT the case
https://github.com/nmfs-ost/asar/blob/68bbe7bddd5dd9bb53443a94141ffefbe2229e3c/R/create_citation.R#L146

Andrea-Havron-NOAA:
# What is the feature?
* Fixes extra `.` in citation for report
* Addresses issue #494 

* Removes extra `.` when `region_specific_part` is set using the default option

* No

### DIFF
--- a/R/create_citation.R
+++ b/R/create_citation.R
@@ -130,7 +130,7 @@ create_citation <- function(
       {
         # Default
         paste0(
-          ". National Marine Fisheries Service, ",
+          "National Marine Fisheries Service, ",
           "[CITY], [STATE]. "
         )
       }


Schiano-NOAA (approved):
LGTM! Thank you so much for the help. I was having such a hard time finding this period. Seems like a new set of eyes was just what this issue needed 😄

PR #530, opened 2026-08-11
[fix]: Fix issue preventing italicized text in rendered skeleton PDF

### LINKED ISSUE #529: [Bug]: font prevents italics in report
### Describe the bug

An {asar} user identified that there was no method to add italics to the report. When they rendered each child doc individually, the italics worked, but not for the document as a whole. @sbreitbart-NOAA identified that the issue is stemming from `mainfont: 'Latin Modern Sans` in the skeleton yaml.

We need to:
- [ ] remove this from the document for now,
- [ ] identify an alternative font that allows for italics, or
- [ ] determine why this font is not allowing italics to be processed.

1. `create_template()`
2. Fill in template with any of the following methods for using italics:
```
(*species name*)
(_species name_)
*(species name)*
*species name*
*species* *name*
_species name_
_species_
*species*
```
3. Render document from the skeleton

### Screenshots

_No response_

_No response_

_No response_

### Additional Context

_No response_

sbreitbart-NOAA:
# What is the feature?
* Remove code setting font from YAML and place in in-header.tex instead
* reconfigure latter file to allow italicized, bold, and ital-bold font to show up in rendered (and tagged) PDFs
* New report text is all sans serif (which I believe was the intent, if we were setting the main font as Latin Modern Sans), like this:

image

image

* see above. Addresses https://github.com/nmfs-ost/asar/issues/529

* The pipeline that creates PDFs. Testing shows the PDFs can still be tagged


Schiano-NOAA (approved):
LGTM!

PR #544, opened 2026-08-25
[feat]: add method to upload file to google drive

Schiano-NOAA:
# What is the feature?
* Adding googledrive dependency but push to drive through the package of the json file

* used googledrive function to push up -- users will need to authenticate to google drive on the first use (possibly every 30 days?)

* no

You can find the test push [here](https://drive.google.com/drive/folders/1eZ_2Y38ftFffr2-EhC30ia1N_4Igacvg).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,6 +28,7 @@ Imports:
     forstringr,
     fs,
     glue,
+    googledrive,
     gt,
     purrr,
     stats,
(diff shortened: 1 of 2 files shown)


sbreitbart-NOAA (approved):
We'll have to update the documentation to explain more about the google drive uploading process / overwriting / etc but for now, LGTM!

PR #545, opened 2026-08-27
[feat]: integrate {journals} package bib files as default bibliographies

### LINKED ISSUE #439: [Feature]: link .bib file to {journals} package

The current bib is not as extensive as it should be and needs some updating for references in the current one.

Write a github action that links to the [journals](https://github.com/nmfs-ost/journals) package to write an all-encompassing .bib file for the most commonly referenced journals in fisheries. We will also need to link in our own file with all of the NOAA stock assessment documents in the format:

```
@TechReport{NEFSC_2020_MTA_FALL,
  %author = "{Northeast Fisheries Science Center}",
  author = "NEFSC",
  shortauthor = "NEFSC",
  title = "Fall Management Track Assessments 2020",
  institution = "US Dept Commer, Northeast Fisheries Science Center",
  % address ="166 Water Street, Woods Hole, MA 02543-1026",
  number="22-08",
  %type={Center Reference Document},
  type={Ref. Doc.},
  note="173p.",
  month="",
  pages = "1--173",
  year = "2022",
  eprinttype={noaair},
  eprint={39404}
  % url="https://repository.library.noaa.gov/view/noaa/39404"
}
```

No -- don't want to keep as is

### Additional context

_No response_

Schiano-NOAA:
# What is the feature?
* change defaults for bib file and create link to {journals} package addressing #439 
* remove the previous `asar_references.bib`
* use the journals package to call the bib files from there (will call the assessment reports when that is done as well)

* used the `journals::download` function to call all bib files then add them to the bibliography section

* no


R/create_template.R
```diff
@@ -89,10 +89,11 @@
 #'
 #' Default: NULL
 #'
-#' @param bib_file File path to bibliography file (`.bib`) used for citing references in
-#' the report
+#' @param bib_file File path to additional bibliography file (`.bib`) used for citing references in
+#' the report. By default, all bibliogrphy files from the journals package and
```
sbreitbart-NOAA:
```suggestion
#' the report. By default, all bibliography files are sourced from the  \pkg{journals} package and 
```

R/create_template.R
```diff
@@ -89,10 +89,11 @@
 #'
 #' Default: NULL
 #'
-#' @param bib_file File path to bibliography file (`.bib`) used for citing references in
-#' the report
+#' @param bib_file File path to additional bibliography file (`.bib`) used for citing references in
+#' the report. By default, all bibliogrphy files from the journals package and 
+#' references for all NMFS stock assessment reports are provided.
```
sbreitbart-NOAA: In which bib file are the SARs? Adding a SAR bib file could be very helpful as the only bib files I see seem to be from journals. Also, I recommend formatting the bib file names so that the titles are easier to understand and more closely reflect the journal titles. Example: change "fish.bib" to "fishery_bulletin.bib", "canjfishaquatsci.bib" to "can-j-fish-aquat-sci.bib", etc.

R/create_template.R
```diff
@@ -89,10 +89,11 @@
 #'
 #' Default: NULL
 #'
-#' @param bib_file File path to bibliography file (`.bib`) used for citing references in
-#' the report
+#' @param bib_file File path to additional bibliography file (`.bib`) used for citing references in
```
sbreitbart-NOAA: I suggest adding info to a vignette about this new workflow. Examples: telling users what to do if they want to add their own bib file; that the files are large and must be opened outside of R (too big); etc

R/create_template.R
```diff
@@ -89,10 +89,11 @@
...
+#' the report. By default, all bibliogrphy files from the journals package and 
+#' references for all NMFS stock assessment reports are provided.
 #'
-#' Default: "asar_references.bib"
+#' Default: NULL
```
sbreitbart-NOAA, line 100: I ran "create_template(bib_file = "test.bib") and it placed it in my working directory, rather than in the "bibliography_files/test.bib" location, as suggested in the skeleton YAML

DESCRIPTION
```diff
@@ -28,6 +28,7 @@ Imports:
     fs,
     glue,
     gt,
+    journals,
```
sbreitbart-NOAA, line 31: Does this need to be in here twice? (here and in remotes, below)

R/create_template.R
Schiano-NOAA: I do not set the names of any of the bib files. They are coming directly from the journals package. I am not sure if there is a specific way they need to be labelled or not.

The SARs won't be added until we make the connection with SIS, but it will be downloaded the same as the others when that feature is available.

R/create_template.R
Schiano-NOAA: Ah okay they I will need to change the path! Good catch

DESCRIPTION
Schiano-NOAA: I am not sure tbh

sbreitbart-NOAA (commented):
Nice job! I highlighted a few places that need changes or could be clarified, but most of this is working as expected. Overall, I think the documentation/workflow should be as as straightforward for the user as possible, and I aimed for my comments to revolve around that idea.

R/create_template.R
kellijohnson-NOAA: Some of the names are collections of journals, such as fish.bib is made up of fishbull.bib, fisheries.bib, fishes.bib, and  fishfish.bib. These names of individual journals and single files like canjfishaquatsci are made from the formal abbreviations of the journals. I do not set these names of the journals, my collaborator at the University of Utah does by matching them to other standard practices. I did set the names of the summary files and I can change those if needed.

R/create_template.R
kellijohnson-NOAA: Maybe that information should be in {journals} and referenced here rather than just here? That way it will help all users of {journals}.

DESCRIPTION
kellijohnson-NOAA: Yes, you have to specify if you are importing or suggesting, and then for all packages not on CRAN you have to specify where they come from in the remotes section.

R/create_template.R
Schiano-NOAA: Thanks Kelli that is super helpful to know. I am okay with keeping as is and just adding documentation somewhere explaining that. @sbreitbart-NOAA thoughts?

R/create_template.R
Schiano-NOAA: Very good point! I am not sure what the recommended method of opening the files would be if the user wants to but I could open a PR to the README? documenting this.

sbreitbart-NOAA: Yes, I think that would be a great addition to {journals} (another topic: which journals are covered in which bib files). Some info would still be helpful to have in asar, though- like how someone would add their own bib file (i.e., specifying the `create_template(bib_file)` arg OR by directly modifying the YAML), and then a link to the new resources in journals mentioned here

R/create_template.R
sbreitbart-NOAA: @kellijohnson-NOAA - by summary files, do you mean the bib files with multiple journals in them? as opposed to a single-journal file like canjfishaquatsci? Changing the names could be a way to clarify their contents. Another idea is to add a README file that explains it in the same folder as the bib files- maybe that's what you're thinking, @Schiano-NOAA ?

R/create_template.R
sbreitbart-NOAA: > Very good point! I am not sure what the recommended method of opening the files would be if the user wants to but I could open a PR to the README? documenting this.

Notepad was fine for me, but SA authors might have other tools they prefer

R/create_template.R
kellijohnson-NOAA: Yes @sbreitbart-NOAA. I can add the file with the "mapping" table of which journals go in their own file and which journals are grouped to the journals package but I don't think it should be downloaded. If you think some explanation is necessary I could add a comment at the top of the file that indicates which journals are included.

sbreitbart-NOAA: Sure, that's a fair solution. I think we should avoid the outcome where users are forced to open the individual (large) files and search for citations each time they're unsure where it lives, so having a mapping table available to consult would solve that issue. Then we could easily link to it in the asar documentation (vignette). Also in favor of adding comments to the tops of the files indicating the journals represented 👍

Schiano-NOAA: @sbreitbart-NOAA I was not thinking of adding a README file. It would be helpful, but do you think it would also just be another file in the user's folder? Maybe documentation in `asar` would be better

sbreitbart-NOAA: I'm ok with nixing the readme idea and just putting the info in an asar vignette. Actually, I don't see anywhere where we discuss how to use the bibliography and add references to reports, so there seems to be an opportunity to add all this to the package! Could probably copy/paste most of what's in Day 1 of the workshop, then add the info relevant to the bib files

sbreitbart-NOAA: My bad, it's actually discussed here: https://nmfs-ost.github.io/asar/articles/markdown_notation.html#references

Schiano-NOAA:
@sbreitbart-NOAA I added some information into the bib_name documentation along with making updates to the bib portion in the markdown vignette. I am also opening a PR with journals to add a blurb in their README regarding some of the things you brought up. Does this sound good to you?

sbreitbart-NOAA:
> @sbreitbart-NOAA I added some information into the bib_name documentation along with making updates to the bib portion in the markdown vignette. I am also opening a PR with journals to add a blurb in their README regarding some of the things you brought up. Does this sound good to you?

Thanks, sounds great! There seem to be a couple remaining errors, though:
* I just did some quick testing and see that `create_template(bib_file = "test.bib")` still puts test.bib in my wd but says it's in the bibliography_files folder in the YAML.
* Also, specifying bib_file prevents the bibliography_files folder and its contents from downloading, so there's no bibliography_files folder to put test.bib inside

Schiano-NOAA:
@sbreitbart-NOAA I'll see why the bib file isn't getting put into the folder as it should.

In regards to the second point, by identifying a bib file, the ones from journals won't be downloaded. I felt like this might be preference, but would it be better to have bib_file as in addition to rather than instead of the {journals} bib files?

sbreitbart-NOAA:
My hunch is that, yes, downloading the journals bib files should happen even if a bib file is specified. People could delete them if they didn't want them but it would be harder for them to download them manually

kellijohnson-NOAA:
If you don't want to download them when a user specifies their own .bib file maybe you can just provide a message saying how to download them?

Schiano-NOAA:
The current documentation for bib_file makes it seem it's in addition to and I'd rather them have access and not use them than have to remember which additional step to perform. @kellijohnson-NOAA 's suggestion is a good one too.

I am making edits to have bib_file be in addition to

Schiano-NOAA:
@sbreitbart-NOAA should be good now!

sbreitbart-NOAA:
@Schiano-NOAA Almost there: the skeleton doesn't render because the path to the bib file isn't relative to the skeleton, as it expects. I think making a bibliography_files folder for everyone, and putting it in there, would simplify this step (and that may be what you're in the middle of doing!)

Schiano-NOAA:
It was supposed to. I have another line in there initially but the `bibliography_files` folder was getting repeated then I took it out and no longer had the issue. I will test again

Schiano-NOAA:
@sbreitbart-NOAA could you screen shot what your folder looks like and the yaml with the bibs so I know for sure what you're describing?

sbreitbart-NOAA:
@Schiano-NOAA Now it's downloading the bib files even if I specify `create_template(bib_file = ...)` 👍 But it's still placing it where the path is (i.e., not inside bibliography_files, as it suggests in the YAML):

image

image

Schiano-NOAA:
@sbreitbart-NOAA what are you using as the test.bib? Is it a real file?

PR #554, opened 2026-09-03
[fix] Fix issue with tests when interactive = F

sbreitbart-NOAA:
# What is the feature?
*  Fix issue with tests when interactive = F

* Allowing separate processing for figure and table names in create_template()

* No


sbreitbart-NOAA:
@Schiano-NOAA I ended up skipping the tests on CI because they pass when tested locally but not remotely. I tried several options and none worked. Basically, the asar functions work fine; it's the steps involved in testing (renaming 05_figures.qmd to 06_figures.qmd, etc.) that fail in the GH testing environment

Schiano-NOAA (approved):
Sounds good and this LGTM! Interesting you can't rename in the GHA testing environment

PR #564, opened 2026-09-15
remove default journals from bib 

BenWilliams-NOAA:
# What is the feature?
Remove the default of loading all of the journal bibs.
 It is a hassle to load a substantial amount of bib entries which will ultimately be ignored, and the files are rather large.
Additionally, many people have their own bib management system that allows for more efficient indexing.

Made a switch, with a default of NULL/FALSE. 
 - If a .bib file is provided the .bib is loaded along with a asar citation bib,
 - If is left NULL, generates asar citation bib
 - If TRUE, loads journal citations and creates asar citation bib

Nope


Schiano-NOAA:
Thank you @BenWilliams-NOAA for your PR and this adjustment to the bibliography files! Totally makes sense if users want to use their own bib systems and be able to opt out of using {journals}. I have reviewed and will be merging this PR shortly. I made a few adjustments to the code and ensured tests are passing which will be merged into main right after this one.

Thanks again. Identifying the issue along with bringing a proposed solution here is greatly appreciated and the support it provides can not be understated 😄

Schiano-NOAA (approved):
Looks good to me! Tests will be fixed in another PR

